PR #3, opened 2015-11-19
POC for Spring Cloud Task

cppwfs:
Basic Cleanup


README.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+== Requirements:
+
+* Java 7 or Above
+
+== Build:
+
+[source,shell,indent=2]
+----
+$ mvn clean install
```
ericbottard: Why clean? Why install ?
What about `mvn package` ? I know we may be promoting clean install elsewhere, but there is no reason, really

pom.xml
```diff
@@ -0,0 +1,54 @@
...
+		<version>1.3.0.RELEASE</version>
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>org.springframework.cloud</groupId>
+	<artifactId>spring-cloud-task</artifactId>
+	<packaging>jar</packaging>
+	<name>task</name>
+	<version>1.0.0.BUILD-SNAPSHOT</version>
+	<description>Spring Cloud Stream task</description>
```
ericbottard: Do we want the word stream here?

pom.xml
```diff
@@ -0,0 +1,54 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.projectlombok</groupId>
+			<artifactId>lombok</artifactId>
+			<version>1.16.6</version>
+			<scope>provided</scope>
+		</dependency>
+	</dependencies>
+
+	<build>
```
ericbottard: Aren't all those configured from some parent already? Should look similar to spring-cloud-stream, shouldn't it?

src/main/java/org/springframework/cloud/task/annotation/Task.java
```diff
@@ -0,0 +1,49 @@
...
+@Documented
+@Inherited
+@Configuration
+@Component
+@Import({ DefaultTaskConfigurer.class })
+public @interface Task {
+
+	public String taskName() default "";
+
+	Class<?>[] value() default {};
```
ericbottard: Please document

src/main/java/org/springframework/cloud/task/config/TaskConfigurer.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Provides a strategy interface for providing configuration
+ * customization to the task system.
+ *
+ * @author Glenn Renfro
+ */
+public interface TaskConfigurer {
+	/**
+	 * Factory method for a {@link TaskRepository}
```
ericbottard: Wrong javadoc

ericbottard: If those are factory methods, it *may* be better not to call them like getters

src/main/java/org/springframework/cloud/task/config/TaskHandler.java
```diff
@@ -0,0 +1,117 @@
...
+	 * and calls repository implementation to store the exit of the task in the repo after
+	 * run returns result.
+	 *
+	 * @param joinPoint
+	 */
+	@AfterReturning("within( @org.springframework.cloud.task.annotation.Task *) && (execution(* org.springframework.boot.CommandLineRunner.run(..)) || execution(* org.springframework.boot.ApplicationRunner.run(..)))")
+	public void afterReturnCommandLineRunner(JoinPoint joinPoint) {
+		int result = 0;
+		List<ExitCodeGenerator> generators = new ArrayList<ExitCodeGenerator>();
+		generators
```
ericbottard: why not inject?

src/main/java/org/springframework/cloud/task/repository/LoggerTaskRepository.java
```diff
@@ -0,0 +1,40 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+/**
+ * @author Glenn Renfro
+ */
+@Component
+public class LoggerTaskRepository implements TaskRepository {
+	Logger logger = LoggerFactory.getLogger(LoggerTaskRepository.class);
```
ericbottard: can make private and final, if not static

src/main/java/org/springframework/cloud/task/repository/TaskExecution.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Represents the state of the Task for each execution
+ *
+ * @author Glenn Renfro
+ */
+@AllArgsConstructor
+@NoArgsConstructor
+@ToString
+public
+@Data
```
ericbottard: weird placement

src/main/java/org/springframework/cloud/task/repository/TaskExecution.java
```diff
@@ -0,0 +1,79 @@
...
+class TaskExecution {
+
+	/**
+	 * The unique id  associated with the task execution
```
ericbottard: dots at end of all javadoc in this class

README.adoc
cppwfs: Great question.  In short the task annotation will be applied to Spring Boot Applications (CommandLineRunner/ApplicationRunners).  Thus we need to have the users of the annotation pull this this projects artifacts into their project as a dependency in a POM file.

pom.xml
cppwfs: Good catch

README.adoc
```diff
@@ -0,0 +1,32 @@
...
+----
+
+== Example:
+
+[source,java,indent=2]
+----
+@Task(taskName = "imSampleB")
```
mminella: If not in this PR, ok, but can we change this to be `@Task("imSampleB")`?

pom.xml
mminella: +1, this should be Spring Cloud Task (capital T)

pom.xml
```diff
@@ -0,0 +1,54 @@
...
+	<groupId>org.springframework.cloud</groupId>
+	<artifactId>spring-cloud-task</artifactId>
+	<packaging>jar</packaging>
+	<name>task</name>
+	<version>1.0.0.BUILD-SNAPSHOT</version>
+	<description>Spring Cloud Stream task</description>
+
+	<properties>
+		<start-class>org.springframework.cloud.task.module.sample.SampleTaskApplication</start-class>
+	</properties>
```
mminella: What is this property for?

src/main/java/org/springframework/cloud/task/annotation/Task.java
```diff
@@ -0,0 +1,49 @@
...
+ * Annotation that identifies a class as a task.
+ *
+ * @author Glenn Renfro
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@Configuration
+@Component
```
mminella: `@Configuration` and `@Component`?

src/main/java/org/springframework/cloud/task/annotation/Task.java
```diff
@@ -0,0 +1,49 @@
...
+@Import({ DefaultTaskConfigurer.class })
```
mminella: I don't think this will work with the plans we have for allowing this to be extended.  Take a look at how `@EnableBatchProcessing` works.  That uses an `ImportSelector` to choose what to import.

src/main/java/org/springframework/cloud/task/config/DefaultTaskConfigurer.java
```diff
@@ -0,0 +1,44 @@
...
+package org.springframework.cloud.task.config;
+
+import org.springframework.cloud.task.repository.LoggerTaskRepository;
+import org.springframework.cloud.task.repository.TaskRepository;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
+
+/**
+ * If no TaskConfigurer is present this configuration will be used.
```
mminella: I don't think that selection works with the way the annotation is coded (feel free to tell me I'm wrong)...

src/main/java/org/springframework/cloud/task/config/DefaultTaskConfigurer.java
```diff
@@ -0,0 +1,44 @@
...
+ * @author Glenn Renfro
+ */
+@Configuration
+public class DefaultTaskConfigurer {
+
+	@Bean
+	@Scope("prototype")
+	public TaskHandler getTaskHandler() {
```
mminella: Annotating these methods is not correct.  That would require everyone to annotate their methods as well as it prevents us from really overriding this configuration.  What we really want to do is import a configuration that uses the `TaskConfigurer` to provide the bean implementations.  Take a look at `AbstractBatchConfiguration` in batch as an example.

src/main/java/org/springframework/cloud/task/config/TaskHandler.java
```diff
@@ -0,0 +1,117 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Offers the advice on how to record tasks to the repository for both applicationrunner
+ * and commandlinerunner spring boot applications.
+ *
+ * @author Glenn Renfro
+ */
+@Aspect
+@Component
```
mminella: Is `@Component` needed here?

src/main/java/org/springframework/cloud/task/repository/LoggerTaskRepository.java
```diff
@@ -0,0 +1,40 @@
...
+
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+/**
+ * @author Glenn Renfro
+ */
+@Component
```
mminella: This probably shouldn't be a `@Component`

src/main/java/org/springframework/cloud/task/repository/TaskExecution.java
mminella: Odd formatting in general.  Also, while nice for prototyping, I'd suggest we not require lombok as a dependency as of right now.

markfisher:
shouldn't there be an `@EnableTasks` that brings in the `TaskConfigurer` so that beans are added to the context (as well as the detection and any necessary registration/configuration of the `@Task` components)?

src/main/java/org/springframework/cloud/task/repository/TaskExecution.java
spencergibb: Why no lombok?

mminella: If it was already being brought in within SCDF, I probably wouldn't argue but to include that dependency for just this one class feels a bit heavy IMHO.

markfisher: -1 to Lombok

pom.xml
```diff
@@ -0,0 +1,42 @@
...
+		<module>spring-cloud-task</module>
+	</modules>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.springframework.boot</groupId>
+				<artifactId>spring-boot-maven-plugin</artifactId>
+				<version>1.3.0.RELEASE</version>
+			</plugin>
```
mminella: I don't think this is correct.  I can fix it on merge if you agree, but I don't think we want a boot über jar for this library...

mminella:
Merged as bd1243cae6d78a3b680d40a235048cb8a2e7cc36.  Further clean up of the configuration options will be done and tracked via additional stories.

PR #15, opened 2015-11-20
Settings file is now present

cppwfs:
Used the same settings file from spring cloud stream

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -13,7 +13,7 @@ process persists beyond the life of the task for future reporting.
 
 [source,shell,indent=2]
 ----
-$ mvn clean install
+$ mvn -s settings.xml clean install
 ----
 
 == Example:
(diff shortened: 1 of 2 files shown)


trisberg:
Should we also add a <profile>  in the pom like we have in spring-cloud-stream? If we have the profile do we even need .settings.xml?

mminella:
It's interesting that spring-cloud-stream has both a profile in the pom.xml and the .settings.xml.  And I thought that having repositories in poms that are published to central are a no-no: http://blog.sonatype.com/2009/02/why-putting-repositories-in-your-poms-is-a-bad-idea/#.Vk-fp9-rSV4 ?

cppwfs:
+1 on Michaels comments.

mbogoevici:
The use of the profile prevents the repositories to show up in the published pom. Although it doesn't feel right. I would advise not to use the profile at all - we should probably get rid of it too ... 
 
http://repo.spring.io/libs-snapshot/org/springframework/cloud/spring-cloud-stream-parent/1.0.0.M2/

Issue #6, opened 2015-11-20
Add TaskRepository implementation to keep track of task executions


sabbyanandan:
As a user, I'd want an out of the box `TaskRepository` implementation that logs task execution status updates.

Issue #5, opened 2015-11-20
Persist task execution statuses in database


sabbyanandan:
As a user, I'd want the ability to store and update task execution statuses in database, so I can use the persisted events for later analysis.

PR #24, opened 2015-12-03
SCT-6 Support RDBMS & Map Repositories

cppwfs:
DefaultTaskConfigurer returns only one SimpleTaskRepository all others have been removed
* SimpleTaskRepository can use JdbcTaskExecutionDao or a MapTaskExecutionDao
* Added Jdbc and Map TaskRepositoryFactoryBeans to support the creation of the correct type of SimpleTaskRepository
* TaskDatabaseInitializer will initialize a database if a datasource is found and if they have not disabled it using the spring.task.initialize.enable
* Started adding tests

This resolves spring-cloud/spring-cloud-task#6 and resolves spring-cloud/spring-cloud-task#5


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/DefaultTaskConfigurer.java
```diff
@@ -16,27 +16,57 @@
...
+					new MapTaskRepositoryFactoryBean();
+			taskRepository = mapTaskRepositoryFactoryBean.getObject();
+		}
+		else {
+			JdbcTaskRepositoryFactoryBean jdbcTaskRepositoryFactoryBean =
+					new JdbcTaskRepositoryFactoryBean(dataSource);
+			taskRepository = jdbcTaskRepositoryFactoryBean.getObject();
+		}
+		return taskRepository;
+	}
```
mminella: We should move this to somewhere so it's only executed once instead of every time we call `getTaskRepository`

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/DefaultTaskConfigurer.java
```diff
@@ -16,27 +16,57 @@
...
+import org.springframework.cloud.task.repository.support.SimpleTaskRepository;
 
 /**
  * If no {@link TaskConfigurer} is present, then this configuration will be used.
  * The following defaults will be used:
- *
  * <ul>
- * <li>{@link LoggerTaskRepository} will be the default {@link TaskRepository}.</li>
+ * <li>{@link SimpleTaskRepository} is the default {@link TaskRepository} returned.
+ * if a data source is present then a data will be stored in the database {@link JdbcTaskExecutionDao} else it will
```
mminella: Capital I

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskHandler.java
```diff
@@ -78,6 +78,7 @@ public void beforeCommandLineRunner(JoinPoint joinPoint) {
 		taskExecution.setTaskName(taskName);
 		taskExecution.setStartTime(new Date());
 		taskExecution.setExecutionId(executionId);
+		taskExecution.setParameters(new ArrayList<String>());
```
mminella: That doesn't look right...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -0,0 +1,162 @@
...
+		this.tablePrefix = tablePrefix;
+	}
+
+	public void setJdbcTemplate(JdbcOperations jdbcTemplate) {
+		this.jdbcTemplate = jdbcTemplate;
+	}
+
+	protected JdbcOperations getJdbcTemplate() {
+		return jdbcTemplate;
+	}
```
mminella: Is there a reason for this to be exposed at all?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -0,0 +1,162 @@
...
+			+ "where TASK_EXECUTION_ID = ?";
+
+
+	private static final String DEFAULT_TABLE_PREFIX = "";
+
+	private String tablePrefix = DEFAULT_TABLE_PREFIX;
+
+	private JdbcOperations jdbcTemplate;
+
+	public JdbcTaskExecutionDao(JdbcOperations jdbcTemplate) {
```
mminella: Should this take a `JdbcOperations` or just a `DataSource`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -0,0 +1,162 @@
...
+	 */
+	private void validateTaskExecution(TaskExecution taskExecution) {
+		Assert.notNull(taskExecution, "taskExecution should not be null");
+		Assert.notNull(taskExecution.getExecutionId(), "taskExecutionId should not be null");
+		Assert.notNull(taskExecution.getStartTime(), "TaskExecution start time cannot be null.");
+	}
+
+	private String getQuery(String base) {
+		return StringUtils.replace(base, "%PREFIX%", tablePrefix);
+	}
```
mminella: We should think about a way to do this only once on creation instead of every time we request the query

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -0,0 +1,50 @@
...
+ * Stores Task Execution Information to a in memory map.
+ *
+ * @author Glenn Renfro
+ */
+public class MapTaskExecutionDao implements TaskExecutionDao {
+
+	private Map<String, TaskExecution> taskExecutions;
+
+	public MapTaskExecutionDao() {
+		taskExecutions = new HashMap<String, TaskExecution>();
```
mminella: `HashMap` isn't thread safe...we'll need something like a `ConcurrentHashMap` for this.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.cloud.task.repository.TaskExecution;
+
+/**
+ * Stores Task Execution Information to a in memory map.
+ *
+ * @author Glenn Renfro
+ */
+public class MapTaskExecutionDao implements TaskExecutionDao {
+
+	private Map<String, TaskExecution> taskExecutions;
```
mminella: `Map` probably doesn't provide a strong enough contract for this...`ConcurrentMap` maybe?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -0,0 +1,50 @@
...
+		taskExecutions.put(taskExecution.getExecutionId(), taskExecution);
+	}
+
+	@Override
+	public void updateTaskExecution(TaskExecution taskExecution) {
+		taskExecutions.put(taskExecution.getExecutionId(), taskExecution);
+	}
+
+	public Map<String, TaskExecution> getTaskExecutions(){
+		return taskExecutions;
```
mminella: This exposes the underlying data structure which means someone could modify it without using the dao.  This needs to return an unmodifiable copy...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
cppwfs: Agreed.

spring-cloud-task-core/pom.xml
```diff
@@ -58,6 +66,17 @@
...
+		<dependency>
+			<groupId>org.hsqldb</groupId>
+			<artifactId>hsqldb</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-dbcp2</artifactId>
+			<scope>test</scope>
+		</dependency>
```
mminella: This is being nit picky, but since Boot prefers the Tomcat connection pool, we probably should use it as well (in fact, they list DBCP2 as the last option they'd use in their docs here: https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-sql.html).

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
```diff
@@ -40,6 +45,12 @@
 	@Autowired
 	private ApplicationContext context;
 
+	@Autowired(required = false)
+	private Collection<DataSource> dataSources;
+
+	@Autowired
+	ResourceLoader resourceLoader;
```
mminella: Why package private?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/TaskExecution.java
```diff
@@ -28,6 +29,7 @@
 public class TaskExecution {
 
 	public TaskExecution() {
+		parameters = new ArrayList<>();
 	}
 
 	public TaskExecution(String executionId, int exitCode, String taskName,
```
mminella: Can we re-arange this so that the fields are above the constructors?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -0,0 +1,158 @@
...
+	/**
+	 * Validate TaskExecution. At a minimum a startTime.
+	 *
+	 * @param taskExecution the taskExecution to be evaluagted.
+	 */
+	private void validateTaskExecution(TaskExecution taskExecution) {
+		Assert.notNull(taskExecution, "taskExecution should not be null");
+		Assert.notNull(taskExecution.getExecutionId(), "taskExecutionId should not be null");
+		Assert.notNull(taskExecution.getStartTime(), "TaskExecution start time cannot be null.");
+	}
```
mminella: Shouldn't this logic live in the `SimpleTaskRepository` so it's consistent across DAO uses?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/DatabaseType.java
```diff
@@ -0,0 +1,115 @@
...
+
+
+/**
+ * Enum representing a database type, such as DB2 or oracle.  The type also
+ * contains a product name, which is expected to be the same as the product name
+ * provided by the database driver's metadata.
+ *
+ * @author Glenn Renfro
+ */
+public enum DatabaseType {
```
mminella: Am I missing something or is this not actually used except within the tests?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskRepositoryFactoryBean.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.springframework.cloud.task.repository.TaskRepository;
+
+/**
+ * Automates the creation of a {@link SimpleTaskRepository} using DAO implementations
+ * which persist task execution data to a datastore.
+ *
+ * @author Glenn Renfro
+ */
+public interface TaskRepositoryFactoryBean {
```
mminella: Do we really need this interface if we use Spring's `FactoryBean` interface (since generics address the typing)?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/JdbcTaskRepositoryFactoryBean.java
```diff
@@ -0,0 +1,73 @@
...
+
+	public static final String DEFAULT_TABLE_PREFIX = "";
+
+	protected static final Log logger = LogFactory.getLog(JdbcTaskRepositoryFactoryBean.class);
+
+	private DataSource dataSource;
+
+	private String tablePrefix = DEFAULT_TABLE_PREFIX;
+
+	public JdbcTaskRepositoryFactoryBean(DataSource dataSource)  {
```
mminella: No arg constructor as an option would be nice instead of having to pass null here...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/JdbcTaskRepositoryFactoryBean.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Automates the creation of a {@link SimpleTaskRepository} using JDBC DAO implementations
+ * which persist task execution data into the database. Requires the user to describe
+ * what kind of database they are using.
+ *
+ * @author Glenn Renfro
+ */
+public class JdbcTaskRepositoryFactoryBean implements TaskRepositoryFactoryBean{
+
+	public static final String DEFAULT_TABLE_PREFIX = "";
```
mminella: In batch, the default is `BATCH_`.  Do we want to consider `TASK_` for our default?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/MapTaskRepositoryFactoryBean.java
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	/**
+	 * Returns the a simpleTaskRepository that utilizes a MapTaskExecutionDao
+	 * @return instance of task repository.
+	 */
+	public TaskRepository getObject(){
+		return taskRepository;
+	}
+}
```
mminella: Missing new line.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/DatabaseType.java
cppwfs: Its also used in GetDatabaseType.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/JdbcTaskRepositoryFactoryBean.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.sql.DataSource;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.cloud.task.repository.TaskRepository;
+import org.springframework.cloud.task.repository.dao.JdbcTaskExecutionDao;
+import org.springframework.cloud.task.repository.dao.TaskExecutionDao;
+
+/**
+ * Automates the creation of a {@link SimpleTaskRepository} using JDBC DAO implementations
```
mminella: In the javadoc for both this and the map one, I wouldn't mention the daos themselves since that's really an implementation detail.  Just that one utilizes a database as the persistence mechanism and one uses and in memory Map (not for production use).

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/SimpleTaskRepository.java
```diff
@@ -16,28 +16,44 @@
...
+		taskExecutionDao.saveTaskExecution(taskExecution);
 		logger.info("Creating: " + taskExecution.toString());
 	}
+
+
+	/**
+	 * Retrieves the taskExecutionDao associated with this repository.
+	 * @return the taskExecutionDao
+	 */
+	public TaskExecutionDao getTaskExecutionDao() {
```
mminella: Why are we exposing this?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskRepositoryFactoryBean.java
cppwfs: It  can be removed.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializer.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.jdbc.datasource.init.ResourceDatabasePopulator;
+import org.springframework.jdbc.support.MetaDataAccessException;
+
+/**
+ * @author Glenn Renfro
+ */
+
+public class TaskDatabaseInitializer {
+
+	public static final String DEFAULT_TABLE_PREFIX = "";
```
mminella: We should consolidate this constant to one place instead of duplicating it across the initializer and the DAO.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
cppwfs: DOH!

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializer.java
```diff
@@ -0,0 +1,83 @@
...
+
+	private static final String DEFAULT_SCHEMA_LOCATION = "classpath:org/springframework/"
+			+ "cloud/task/schema-@@platform@@.sql";
+
+	@Autowired(required = false)
```
mminella: A datasource is not required for this class?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializer.java
```diff
@@ -0,0 +1,83 @@
...
+	private DataSource dataSource;
+
+	@Autowired
+	private ResourceLoader resourceLoader;
```
mminella: I wouldn't autowire these, especially since we do some analysis in the Configurer on if there is more than one.  Also, isn't the DataSource required for this class?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializer.java
```diff
@@ -0,0 +1,83 @@
...
+	 * Path to the SQL file to use to initialize the database schema.
+	 */
+	private String schema = DEFAULT_SCHEMA_LOCATION;
+
+	private String tablePrefix = DEFAULT_TABLE_PREFIX;
+
+
+	@PostConstruct
+	public void initializeDatabase() {
+		if (dataSource !=null) {
```
mminella: I'd make the DataSource required and kill this if statement...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializer.java
```diff
@@ -0,0 +1,83 @@
...
+			}
+			ResourceDatabasePopulator populator = new ResourceDatabasePopulator();
+			String schemaLocation = schema;
+			schemaLocation = schemaLocation.replace("@@platform@@", platform);
+			populator.addScript(resourceLoader.getResource(schemaLocation));
+			populator.setContinueOnError(false);
+			DatabasePopulatorUtils.execute(populator, this.dataSource);
+		}
+	}
+	private String getDatabaseType() {
```
mminella: White space between methods.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/DatabaseType.java
```diff
@@ -0,0 +1,115 @@
...
+	DB2("DB2"),
+	DB2ZOS("DB2ZOS"),
+	HSQL("HSQL Database Engine"),
+	SQLSERVER("Microsoft SQL Server"),
+	MYSQL("MySQL"),
+	ORACLE("Oracle"),
+	POSTGRES("PostgreSQL"),
+	SYBASE("Sybase"),
+	H2("H2"),
+	SQLITE("SQLite");
```
mminella: We have all these values, yet we're only providing the HSQLDB schema...I'd only add the values if we have the schemas for them.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskRepositoryFactoryBean.java
```diff
@@ -0,0 +1,30 @@
...
+
+	public TaskRepository getObject();
+}
```
mminella: New line

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-hsqldb.sql
```diff
@@ -0,0 +1,18 @@
...
+	LAST_UPDATED TIMESTAMP ,
+	STATUS_CODE VARCHAR(10)
+);
+
+CREATE TABLE TASK_EXECUTION_PARAMS  (
+	TASK_EXECUTION_ID VARCHAR(100) NOT NULL ,
+	TASK_PARAM VARCHAR(250) ,
+	constraint TASK_EXEC_PARAMS_FK foreign key (TASK_EXECUTION_ID)
+	references TASK_EXECUTION(TASK_EXECUTION_ID)
+) ;
```
mminella: New line

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/SimpleTaskRepository.java
cppwfs: In this case so that the Tests can retrieve the Map repository.  Also will be used for TaskExplorer when using the MapRepository.

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-hsqldb.sql
```diff
@@ -0,0 +1,18 @@
+
+CREATE TABLE TASK_EXECUTION  (
+	TASK_EXECUTION_ID VARCHAR(100)  NOT NULL PRIMARY KEY ,
```
mminella: Why is this a varchar and not a long?

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-hsqldb.sql
```diff
@@ -0,0 +1,18 @@
...
+	END_TIME TIMESTAMP DEFAULT NULL ,
+	TASK_NAME  VARCHAR(100) ,
+	EXIT_CODE INTEGER ,
+	EXIT_MESSAGE VARCHAR(2500) ,
+	LAST_UPDATED TIMESTAMP ,
+	STATUS_CODE VARCHAR(10)
+);
+
+CREATE TABLE TASK_EXECUTION_PARAMS  (
+	TASK_EXECUTION_ID VARCHAR(100) NOT NULL ,
```
mminella: Why is this a varchar and not a long?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/DatabaseType.java
cppwfs: So Postgres and HSQLDB to start?

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-hsqldb.sql
cppwfs: TaskExecutionID is a UUID String.

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-hsqldb.sql
cppwfs: TaskExecutionID is a UUID String.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDaoTests.java
```diff
@@ -0,0 +1,71 @@
...
+		JdbcTaskExecutionDao dao = new JdbcTaskExecutionDao(db);
+		TaskExecution expectedTaskExecution = TestUtils.createSampleTaskExecutionNoParam();
+		dao.saveTaskExecution(expectedTaskExecution);
+
+		TestUtils.verifyTaskExecution(expectedTaskExecution,
+				TestUtils.getTaskExecutionFromDB(db, expectedTaskExecution.getExecutionId()));
+	}
+
+	@Test
+	public void updateTaskExecution(){
```
mminella: How about an update test for a TaskExecution that does not exist (aka hasn't been added yet)?

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDaoTests.java
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+
+	@After
+	public void tearDown() {
+		db.shutdown();
+	}
+
+	@Test
+	public void saveTaskExecution(){
```
mminella: What about a test for saving a dupe?

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDaoTests.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public class JdbcTaskExecutionDaoTests {
+
+	private EmbeddedDatabase db;
+
+	@Before
+	public void setup() {
+		db = new EmbeddedDatabaseBuilder()
+				.generateUniqueName(true)
+				.addScripts("task-schema.sql")
```
mminella: Why do we have the schema in here twice?  Why not just use the official one?

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDaoTests.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.junit.Test;
+import org.springframework.cloud.task.repository.TaskExecution;
+import org.springframework.cloud.task.util.TestUtils;
+
+/**
+ * Executes unit tests on MapTaskExecutionDaoTests.
+ * @author Glenn Renfro
+ */
+public class MapTaskExecutionDaoTests {
```
mminella: Same notes on tests that I made for the Jdbc version of the dao.  In fact, since the DAO's should behave the same, you could combine the test code into a super class and just add the initialization code in the child test classes...

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/support/SimpleTaskRepositoryJdbcTests.java
```diff
@@ -0,0 +1,87 @@
...
+
+	private EmbeddedDatabase db;
+
+	private TaskRepository taskRepository;
+
+	@Before
+	public void setUp() {
+		db = new EmbeddedDatabaseBuilder()
+				.generateUniqueName(true)
+				.addScripts("task-schema.sql")
```
mminella: Same note about the schema...this should use the official one, not a dupe in the test dir.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/util/TestUtils.java
```diff
@@ -0,0 +1,228 @@
...
+		return expectedTaskExecution;
+	}
+
+	/**
+	 * Updates a sample TaskExecution in the taskRepository.
+	 * @param taskRepository the taskRepository where the taskExecution should be updated.
+	 * @return the taskExecution created.
+	 */
+	public static TaskExecution updateTaskExecution(TaskRepository taskRepository,
+																	String taskExecutionId){
```
mminella: Odd indentation

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/util/TestUtils.java
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	/**
+	 * Retrieves the TaskExecution from the datasource
+	 * @param dataSource The datasource from which to retrieve the taskExecution
+	 * @param taskExecutionId The id of the task to search .
+	 * @return taskExecution
+	 */
+	public static TaskExecution getTaskExecutionFromDB(DataSource dataSource,
+													   String taskExecutionId){
```
mminella: Odd indentation

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/util/TestUtils.java
```diff
@@ -0,0 +1,228 @@
...
+		TaskExecution taskExecution = new TaskExecution();
+		for (Map row : rows) {
+			taskExecution.setExecutionId((String)row.get("TASK_EXECUTION_ID"));
+			taskExecution.setStartTime((Date) row.get("START_TIME"));
+			taskExecution.setEndTime((Date) row.get("END_TIME"));
+			taskExecution.setExitCode((Integer)row.get("EXIT_CODE"));
+			taskExecution.setExitMessage((String)row.get("EXIT_MESSAGE"));
+			taskExecution.setStatusCode((String)row.get("STATUS_CODE"));
+			taskExecution.setTaskName((String)row.get("TASK_NAME"));
+		}
```
mminella: `RowMapper`?

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/util/TestUtils.java
```diff
@@ -0,0 +1,228 @@
...
+import org.springframework.cloud.task.repository.TaskExecution;
+import org.springframework.cloud.task.repository.TaskRepository;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+/**
+ * Offers utils to test the log results produced by the code being tested.
+ *
+ * @author Glenn Renfro
+ */
+public class TestUtils {
```
mminella: This feels like a dumping ground for various test stuff.  The javadoc doesn't reflect what this class actually is and it probably should be broken up.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/support/DatabaseTypeTests.java
```diff
@@ -0,0 +1,130 @@
...
+import javax.sql.DataSource;
+
+import org.junit.Test;
+import org.springframework.cloud.task.util.DatabaseTypeTestUtils;
+import org.springframework.jdbc.support.MetaDataAccessException;
+
+/**
+ * Tests that the correct database names are selected from datasource metadata.
+ *
+ * @author Lucas Ward
```
mminella: Ha...He'll get a kick out of this.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializer.java
cppwfs: Agreed we need to test for multiple datasources and we can remove the @Autowired.  We can check for datasources in the context at postconstruct time.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializer.java
cppwfs: We need to have this check because the class tests for a datasource, if present it initializes the database, else it does nothing.  And we can't place it as a conditional because, the datasource may not have been instantiated at the time the TaskDatabaseInitializer is instantiated.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/support/DatabaseTypeTests.java
cppwfs: Well it was his code ;-)

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializerTests.java
```diff
@@ -0,0 +1,108 @@
...
+		assertNotNull(this.context.getBean(SimpleTaskRepository.class));
+		SimpleTaskRepository repository = this.context.getBean(SimpleTaskRepository.class);
+		assertNotNull(repository);
+		assertThat(repository.getTaskExecutionDao(), instanceOf(MapTaskExecutionDao.class));
+		MapTaskExecutionDao dao = (MapTaskExecutionDao) repository.getTaskExecutionDao();
+		assertEquals(0, dao.getTaskExecutions().size());
+	}
+
+	@Test
+	public void testNoBatchConfiguration() throws Exception {
```
mminella: `testNoBatchConfiguration`?  I think you mean, `testNoTaskConfiguration` ;)

spring-cloud-task-core/src/test/resources/task-schema.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE TASK_EXECUTION  (
```
mminella: We should remove this file and just use the official one.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializerTests.java
```diff
@@ -0,0 +1,108 @@
...
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+/**
+ * Verifies that task initialization occurs properly.
+ *
+ * @author Glenn Renfro
+ */
+public class TaskDatabaseInitializerTests {
```
mminella: What about a test for a context with multiple `DataSources`?

mminella:
One other thing to think about that I don't see here...transactionality.  Since we have two tables, they need to remain in sink.  I believe batch uses an aspect to wrap repository interactions with a transaction but I'd have to double check (and there may be a better way to handle that now).

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDaoTests.java
cppwfs: While the code is similar in how we store or update the tasks in that the DAO's do behave the same way.  Retrieving the actual results differ between the 2 tests i.e. the relational database and the map.

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-hsqldb.sql
cppwfs: done

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/support/TaskDatabaseInitializerTests.java
cppwfs: done

ghillert:
@cppwfs Please double-check the JavaDoc comments, saw a couple of unnecessary closing tags:

http://blog.joda.org/2012/11/javadoc-coding-standards.html

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/DefaultTaskConfigurer.java
```diff
@@ -16,27 +16,62 @@
...
 	}
 
 	public TaskRepository getTaskRepository() {
-		return new LoggerTaskRepository();
+		return taskRepository;
+	}
+
+	public TaskExplorer getTaskExplorer() {
+		return null;
+		//TODO if datasource != null use TaskRepositoryFactoryBean from above like initialize method in DefaultBatchConfigurer
```
ghillert: Should this return null or rather throw some exception?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
```diff
@@ -34,12 +39,22 @@
...
  */
+@EnableTransactionManagement
 @Configuration
 public class SimpleTaskConfiguration {
 
 	@Autowired
 	private ApplicationContext context;
 
+	@Autowired(required = false)
+	private Collection<DataSource> dataSources;
```
ghillert: Should we provide setter for the `DataSource` so users can provide their own? E.g. you have more than one, but you can provide the *right* one, so you can still use the class.

ghillert: E.g. should maybe `@EnableTask` accept a `dataSource` property?

ghillert:
@cppwfs Ensure that all packages contain proper `package-info.java` files.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/DatabaseType.java
```diff
@@ -0,0 +1,107 @@
...
+		}
+		else{
+			return nameMap.get(productName);
+		}
+	}
+
+	private String getProductName() {
+		return productName;
+	}
+
```
ghillert: remove extra lines

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/DatabaseType.java
```diff
@@ -0,0 +1,107 @@
...
+	POSTGRES("PostgreSQL");
+
+	private static final Map<String, DatabaseType> nameMap;
+
+	private final String productName;
+
+	private DatabaseType(String productName) {
+		this.productName = productName;
+	}
+
```
ghillert: remove extra line

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/DatabaseType.java
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @author Glenn Renfro
+ */
+public enum DatabaseType {
+
+	HSQL("HSQL Database Engine"),
+	ORACLE("Oracle"),
+	POSTGRES("PostgreSQL");
+
+	private static final Map<String, DatabaseType> nameMap;
```
ghillert: maybe name `nameMap` more appropriately.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import org.springframework.cloud.task.repository.TaskExecution;
+
+/**
+ * Stores Task Execution Information to a in memory map.
```
ghillert: spelling

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
cppwfs: To your first point, if the user wants a setter, then they should probably implement their own configurer.  
To the second point, I'll raise a github issue and let's discuss.

cppwfs: https://github.com/spring-cloud/spring-cloud-task/issues/26

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/DefaultTaskConfigurer.java
```diff
@@ -16,27 +16,62 @@
...
+	}
+
+	private void initialize(){
+		if (dataSource == null) {
```
ghillert: One think I am wondering - maybe adding some debug-level logging statements — e.g. `Creating JdbcTaskRepository` or `Creating MapTaskRepository`, since some magic is involved.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/JdbcTaskRepositoryFactoryBean.java
```diff
@@ -0,0 +1,78 @@
...
+		this.tablePrefix = tablePrefix;
+	}
+
+	/**
+	 * Returns the a simpleTaskRepository that utilizes a MapTaskExecutionDao
+	 * @return instance of task repository.
+	 */
+	public TaskRepository getObject(){
+		TaskRepository taskRepository = null;
+		logger.debug("Creating SimpleTaskRepository that will use a JdbcTaskExecutionDao");
```
ghillert: To make things refactoring-safe use something like `SimpleTaskRepository.class.getName()`, same for JdbcTaskExecutionDao.

ghillert:
Merged with a9f882f24b536002a1b8f183630ee92ec4d39bf2

Issue #7, opened 2015-11-20
Add TaskExplorer implementation


sabbyanandan:
As a user, I'd want the `TaskExplorer` implementation, so I can use this contract to query task details [such as: _name_, _execution-count_ etc.] from the database.

cppwfs:
Needs to support Pageable
i.e. spring data Pageable interface (i.e. spring data commons)

PR #30, opened 2015-12-10
SCT-7 Add TaskExplorer implementation

cppwfs:
* Supports SimpleTaskExplorer that accepts a TaskExecutionDao.
* Updated Dao's to support the explorer query requirements
* Added tests.
* Rebased

This resolves spring-cloud/spring-cloud-task#7

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
@@ -16,15 +16,26 @@
 
 package org.springframework.cloud.task.repository.dao;
 
+import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.sql.Types;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import javax.sql.DataSource;
 
 import org.springframework.cloud.task.repository.TaskExecution;
+import org.springframework.dao.DataAccessException;
+import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.core.JdbcOperations;
 import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.ResultSetExtractor;
+import org.springframework.jdbc.core.RowCallbackHandler;
+import org.springframework.jdbc.core.RowMapper;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
@@ -51,6 +62,30 @@ public class JdbcTaskExecutionDao implements TaskExecutionDao {
 			+ "EXIT_MESSAGE = ?, LAST_UPDATED = ?, STATUS_CODE = ? "
 			+ "where TASK_EXECUTION_ID = ?";
 
+	private static final String GET_EXECUTION_BY_ID = "SELECT TASK_EXECUTION_ID, " +
+			"START_TIME, END_TIME, TASK_NAME, EXIT_CODE, "
+			+ "EXIT_MESSAGE, LAST_UPDATED, STATUS_CODE "
+			+ "from %PREFIX%EXECUTION where TASK_EXECUTION_ID = ?";
+
+	private static final String FIND_PARAMS_FROM_ID = "SELECT TASK_EXECUTION_ID, "
+			+ "TASK_PARAM from %PREFIX%EXECUTION_PARAMS where TASK_EXECUTION_ID = ?";
+
+	private static final String TASK_EXECUTION_COUNT = "SELECT COUNT(*) FROM " +
+			"%PREFIX%EXECUTION where TASK_NAME = ?";
+
+	private static final String FIND_RUNNING_TASK_EXECUTIONS = "SELECT TASK_EXECUTION_ID, "
+			+ "START_TIME, END_TIME, TASK_NAME, EXIT_CODE, "
+			+ "EXIT_MESSAGE, LAST_UPDATED, STATUS_CODE "
+			+ "from %PREFIX%EXECUTION where TASK_NAME = ? AND END_TIME IS NULL "
+			+ "order by TASK_EXECUTION_ID";
+
+	private static final String FIND_TASK_EXECUTIONS = "SELECT TASK_EXECUTION_ID, "
+			+ "START_TIME, END_TIME, TASK_NAME, EXIT_CODE, "
+			+ "EXIT_MESSAGE, LAST_UPDATED, STATUS_CODE "
+			+ "from %PREFIX%EXECUTION where TASK_NAME = ? "
+			+ "order by TASK_EXECUTION_ID";
+
+	final String FIND_TASK_NAMES = "SELECT distinct TASK_NAME from %PREFIX%EXECUTION order by TASK_NAME";
 
 	private static final String DEFAULT_TABLE_PREFIX = "TASK_";
 
@@ -75,7 +110,7 @@ public class JdbcTaskExecutionDao implements TaskExecutionDao {
 				parameters,
 				new int[]{ Types.VARCHAR, Types.TIMESTAMP, Types.TIMESTAMP, Types.VARCHAR,
 						Types.INTEGER, Types.VARCHAR, Types.TIMESTAMP, Types.VARCHAR });
-		insertJobParameters(taskExecution.getExecutionId(), taskExecution.getParameters());
+		insertTaskParameters(taskExecution.getExecutionId(), taskExecution.getParameters());
 	}
 
 	@Override
@@ -109,6 +144,92 @@ public class JdbcTaskExecutionDao implements TaskExecutionDao {
 		this.tablePrefix = tablePrefix;
 	}
 
+	@Override
+	public TaskExecution getTaskExecution(String executionId) {
+		try {
+			TaskExecution taskExecution = jdbcTemplate.queryForObject(getQuery(GET_EXECUTION_BY_ID),
+					new TaskExecutionRowMapper(), executionId);
+			taskExecution.setParameters(getTaskParameters(executionId));
+			return taskExecution;
+		}
+		catch (EmptyResultDataAccessException e) {
+			return null;
+		}
+	}
+
+	@Override
+	public long getTaskExecutionCount(String taskName) {
+		try {
+		return jdbcTemplate.queryForObject(
+				getQuery(TASK_EXECUTION_COUNT), new Object[] { taskName }, Long.class);
+		}
+		catch (EmptyResultDataAccessException e) {
+			return 0;
+		}
+
+	}
+
+	@Override
+	public Set<TaskExecution> findRunningTaskExecutions(String taskName) {
+		final Set<TaskExecution> result = new HashSet<TaskExecution>();
+		RowCallbackHandler handler = new RowCallbackHandler() {
+			@Override
+			public void processRow(ResultSet rs) throws SQLException {
+				TaskExecutionRowMapper mapper = new TaskExecutionRowMapper();
+				result.add(mapper.mapRow(rs, 0));
+			}
+		};
+		jdbcTemplate.query(getQuery(FIND_RUNNING_TASK_EXECUTIONS),
+				new Object[] { taskName }, handler);
+
+		return result;
+
+	}
+
+	@Override
+	public List<TaskExecution> getTaskExecutionsByName(String taskName, final int start,
+													   final int count) {
+		ResultSetExtractor<List<TaskExecution>> extractor =
+				new ResultSetExtractor<List<TaskExecution>>() {
+
+			private List<TaskExecution> list = new ArrayList<TaskExecution>();
+
+			@Override
+			public List<TaskExecution> extractData(ResultSet rs) throws SQLException,
+					DataAccessException {
+				int rowNum = 0;
+				while (rowNum < start && rs.next()) {
+					rowNum++;
+				}
+				while (rowNum < start + count && rs.next()) {
+					RowMapper<TaskExecution> rowMapper = new TaskExecutionRowMapper();
+					list.add(rowMapper.mapRow(rs, rowNum));
+					rowNum++;
+				}
+				return list;
+			}
+
+		};
+
+		List<TaskExecution> result = jdbcTemplate.query(getQuery(FIND_TASK_EXECUTIONS),
+				new Object[] { taskName }, extractor);
+
+		return result;
+	}
+
+	@Override
+	public List<String> getTaskNames() {
+		return jdbcTemplate.query(getQuery(FIND_TASK_NAMES),
+				new RowMapper<String>() {
+					@Override
+					public String mapRow(ResultSet rs, int rowNum)
+							throws SQLException {
+						return rs.getString(1);
+					}
+				});
+	}
+
+
 	private String getQuery(String base) {
 		return StringUtils.replace(base, "%PREFIX%", tablePrefix);
 	}
@@ -120,7 +241,7 @@ public class JdbcTaskExecutionDao implements TaskExecutionDao {
 	 * @param executionId    The executionId to which the params are associated.
 	 * @param taskParameters The parameters to be stored.
 	 */
-	private void insertJobParameters(String executionId, List<String> taskParameters) {
+	private void insertTaskParameters(String executionId, List<String> taskParameters) {
 		for (String param : taskParameters) {
 			insertParameter(executionId, param);
 		}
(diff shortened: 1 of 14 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -139,4 +258,45 @@ private void insertParameter(String executionId, String param) {
...
+	 *
+	 */
+	private final class TaskExecutionRowMapper implements RowMapper<TaskExecution> {
+
+		public TaskExecutionRowMapper() {
+		}
+
+		@Override
+		public TaskExecution mapRow(ResultSet rs, int rowNum) throws SQLException {
+			String  id = rs.getString(1);
```
mminella: Why are we using the index here and not the column name?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -139,4 +258,45 @@ private void insertParameter(String executionId, String param) {
...
+			TaskExecution taskExecution=new TaskExecution();
+			taskExecution.setExecutionId(rs.getString(1));
```
mminella: We just got this above, why get it again?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -139,4 +258,45 @@ private void insertParameter(String executionId, String param) {
 		jdbcTemplate.update(getQuery(CREATE_TASK_PARAMETER), args, argTypes);
 	}
 
+	private List<String> getTaskParameters(String executionId){
```
mminella: We should consider providing a more rich parameters object here instead of just a list of strings...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -139,4 +258,45 @@ private void insertParameter(String executionId, String param) {
...
+
+		jdbcTemplate.query(getQuery(FIND_PARAMS_FROM_ID), new Object[] { executionId },
+				handler);
+		return Collections.unmodifiableList(params);
+
+	}
+	/**
+	 * Re-usable mapper for {@link TaskExecution} instances.
+	 *
+	 * @author Dave Syer
```
mminella: Did Dave really write this?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -109,6 +144,90 @@ public void setTablePrefix(String tablePrefix) {
...
+	@Override
+	public TaskExecution getTaskExecution(String executionId) {
+		try {
+			TaskExecution taskExecution = jdbcTemplate.queryForObject(getQuery(GET_EXECUTION_BY_ID),
+					new TaskExecutionRowMapper(), executionId);
+			taskExecution.setParameters(getTaskParameters(executionId));
+			return taskExecution;
+		}
+		catch (EmptyResultDataAccessException e) {
+			return null;
```
mminella: Hmmmm...Is this an ok condition?  Thoughts on throwing an exception here if the result isn't found?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -109,6 +144,90 @@ public void setTablePrefix(String tablePrefix) {
...
+					list.add(rowMapper.mapRow(rs, rowNum));
+					rowNum++;
+				}
+				return list;
+			}
+
+		};
+
+		List<TaskExecution> result = jdbcTemplate.query(getQuery(FIND_TASK_EXECUTIONS),
+				new Object[] { taskName }, extractor);
```
mminella: Why are we using this extractor instead of a RowMapper?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -109,6 +144,90 @@ public void setTablePrefix(String tablePrefix) {
...
+		final Set<TaskExecution> result = new HashSet<TaskExecution>();
+		RowCallbackHandler handler = new RowCallbackHandler() {
+			@Override
+			public void processRow(ResultSet rs) throws SQLException {
+				TaskExecutionRowMapper mapper = new TaskExecutionRowMapper();
+				result.add(mapper.mapRow(rs, 0));
+			}
+		};
+		jdbcTemplate.query(getQuery(FIND_RUNNING_TASK_EXECUTIONS),
+				new Object[] { taskName }, handler);
```
mminella: Why use a `RowCallbackHandler` here instead of just a `RowMapper` with `queryForObject`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -109,6 +144,90 @@ public void setTablePrefix(String tablePrefix) {
...
+	@Override
+	public List<String> getTaskNames() {
+		return jdbcTemplate.query(getQuery(FIND_TASK_NAMES),
+				new RowMapper<String>() {
+					@Override
+					public String mapRow(ResultSet rs, int rowNum)
+							throws SQLException {
+						return rs.getString(1);
+					}
+				});
```
mminella: Does `queryForObject` and passing the type String not work?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -139,4 +258,45 @@ private void insertParameter(String executionId, String param) {
...
+		RowCallbackHandler handler = new RowCallbackHandler() {
+			@Override
+			public void processRow(ResultSet rs) throws SQLException {
+				params.add(rs.getString(2));
+			}
+		};
+
+		jdbcTemplate.query(getQuery(FIND_PARAMS_FROM_ID), new Object[] { executionId },
+				handler);
+		return Collections.unmodifiableList(params);
```
mminella: We probably don't need to wrap this return object...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -46,7 +51,67 @@ public void updateTaskExecution(TaskExecution taskExecution) {
...
+
+	@Override
+	public long getTaskExecutionCount(String taskName) {
+		int count = 0;
+		for (Map.Entry<String, TaskExecution> entry : taskExecutions.entrySet()) {
+			if (entry.getValue().getTaskName().equals(taskName)) {
+				count++;
+			}
+		}
+		return count;
```
mminella: Code like this makes me sad we can't use Java 8 yet...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -46,7 +51,67 @@ public void updateTaskExecution(TaskExecution taskExecution) {
...
+	@Override
+	public Set<TaskExecution> findRunningTaskExecutions(String taskName) {
+		Set<TaskExecution> result = new HashSet<>();
+		for (Map.Entry<String, TaskExecution> entry : taskExecutions.entrySet()) {
+			if (entry.getValue().getTaskName().equals(taskName) &&
+					entry.getValue().getEndTime() == null) {
+				result.add(entry.getValue());
+			}
+		}
+		return Collections.unmodifiableSet(result);
```
mminella: Do we need to wrap this?  The Set is new for each call to the method so it's not like we're exposing the underlying datastore...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -46,7 +51,67 @@ public void updateTaskExecution(TaskExecution taskExecution) {
...
+	}
+
+	@Override
+	public List<TaskExecution> getTaskExecutionsByName(String taskName, int start, int count) {
+		List<TaskExecution> result = new ArrayList<>();
+		Set<TaskExecution> filteredSet = new HashSet<>();
```
mminella: Does this need to be ordered in some way?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -109,6 +144,90 @@ public void setTablePrefix(String tablePrefix) {
...
+				while (rowNum < start && rs.next()) {
+					rowNum++;
+				}
+				while (rowNum < start + count && rs.next()) {
+					RowMapper<TaskExecution> rowMapper = new TaskExecutionRowMapper();
+					list.add(rowMapper.mapRow(rs, rowNum));
+					rowNum++;
+				}
+				return list;
+			}
```
mminella: Why aren't we using limit clauses here?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -46,7 +51,67 @@ public void updateTaskExecution(TaskExecution taskExecution) {
...
+		int rowNum = 0;
+		Iterator<TaskExecution> rs = filteredSet.iterator();
+		while (rowNum < start && rs.hasNext()) {
+			rs.next();
+			rowNum++;
+		}
+		while (rowNum < start + count && rs.hasNext()) {
+			result.add(rs.next());
+			rowNum++;
+		}
```
mminella: Assuming you use a `SortedSet` the pagination piece becomes just using subset...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -46,7 +51,67 @@ public void updateTaskExecution(TaskExecution taskExecution) {
...
+
+		return Collections.unmodifiableList(result);
```
mminella: Do we need to wrap this given the fact that it's a new data structure for each call?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -46,7 +51,67 @@ public void updateTaskExecution(TaskExecution taskExecution) {
...
+	}
+
+	@Override
+	public List<String> getTaskNames() {
+		Set<String> result = new HashSet<>();
+		for (Map.Entry<String, TaskExecution> entry : taskExecutions.entrySet()) {
+			result.add(entry.getValue().getTaskName());
+		}
+		return Collections.unmodifiableList(new ArrayList(result));
```
mminella: 1.  I doubt we need to wrap this given you're not exposing the underlying data structure.
2.  Why start with a Set only to convert it to a List?  Why not just start with the List?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/JdbcTaskExplorerFactoryBean.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.cloud.task.repository.dao.JdbcTaskExecutionDao;
+import org.springframework.cloud.task.repository.dao.TaskExecutionDao;
+
+/**
+ * Automates the creation of a {@link SimpleTaskExplorer} which will retrieve task
+ * execution data from a database.
+ *
+ * @author Glenn Renfro
+ */
+public class JdbcTaskExplorerFactoryBean {
```
mminella: Why does this not implement `FactoryBean`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/MapTaskExplorerFactoryBean.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.cloud.task.repository.TaskExplorer;
+import org.springframework.cloud.task.repository.dao.MapTaskExecutionDao;
+
+/**
+ * Automates the creation of a {@link SimpleTaskExplorer} which will retrieve task
+ * execution data from a in-memory map.
+ *
+ * @author Glenn Renfro
+ */
+public class MapTaskExplorerFactoryBean {
```
mminella: Why doesn't this implement `FactoryBean`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
cppwfs: In this case we probably want to return null, because the other DAO's will return null if the TaskExecutionId is not found.

mminella: Regardless of what the other DAO returns, the idea of me asking for TaskExecution X and it not being there feels like an error condition to me.  Thoughts?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
cppwfs: Not sure I understand.  We are using a RowMapper to extract the task execution data from each of  rows in the result set.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
cppwfs: Not sure I understand.  We are using the rowmapper for each of rows in the result set.  The handler is here so that we can process each row.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
mminella: You're wrapping a `RowMapper` in a `ResultSetExtractor`.  Why wrap?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
mminella: That's what a RowMapper does...you shouldn't need to wrap it in a RowCallbackHandler.  Am I missing something?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
cppwfs: In this case then we will need to update the SimpleTaskExplorer check for nulls and throw an assert if the result returned from any dao is null.  And catch this exception and return the appropriate exception.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
trisberg: I would try `queryForList` passing in `String.class` as the type -- see http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/jdbc/core/JdbcTemplate.html#queryForList-java.lang.String-java.lang.Class-

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
cppwfs: Agreed let's discuss

mminella:
Merged.

Issue #41, opened 2015-12-11
User wants to support pagination for execution result sets


cppwfs:
for result sets from the Explorer interface should return pageable.

PR #44, opened 2015-12-18
SCT-41 Introduces findAll for task explorer

cppwfs:
resolves spring-cloud/spring-cloud-task#41


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/TaskExplorer.java
```diff
@@ -68,5 +78,12 @@
 	 */
 	public List<TaskExecution> getTaskExecutionsByName(String taskName, int start, int count);
 
+	/**
+	 * Retrieves all the task executions within the pageable constraints.
+	 *
+	 * @param pageable the constraints for the search
+	 * @return page containing the results from the search
+	 */
```
mminella: Any insights into what the default sorting we use is (since we need to sort by something to be reliably pagable) would be useful.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -79,6 +85,12 @@
 	private static final String FIND_TASK_EXECUTIONS = "SELECT TASK_EXECUTION_ID, "
 			+ "START_TIME, END_TIME, TASK_NAME, EXIT_CODE, "
 			+ "EXIT_MESSAGE, LAST_UPDATED, STATUS_CODE "
+			+ "from %PREFIX%EXECUTION order by TASK_EXECUTION_ID "
```
mminella: This will need to be updated to be a sequence number (spring-cloud/spring-cloud-task#34) to truly be useful.  Sorting by a random GUID will be just junk.  At least we should sort by start time for the time being?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -101,6 +109,19 @@ public long getTaskExecutionCount(String taskName) {
 		return new ArrayList<String>(result);
 	}
 
+	@Override
+	public Page<TaskExecution> findAll(Pageable pageable) {
+		Set<TaskExecution> sortedSet = getTaskExecutionTreeSet();
+		sortedSet.addAll(taskExecutions.values());
+		List<TaskExecution> result = new ArrayList<>(sortedSet);
```
mminella: Since order matters to us, why not use `ConcurrentSkipListMap` instead of creating a `TreeSet` every time we ask for the results?  Then you (I believe) should be able to go straight from the `Map` to the `List` without this `Set` step...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/NoOpTaskExplorer.java
```diff
@@ -22,6 +22,9 @@
```
mminella: Do we still need this class if we have the `Map` implementation of the `TaskRepository`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
cppwfs: In this case the key is the executionId.  And we can't use the value of the Map in the Comparator. :-(

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/NoOpTaskExplorer.java
cppwfs: Not really.  I'll get rid of it.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -79,6 +85,12 @@
 	private static final String FIND_TASK_EXECUTIONS = "SELECT TASK_EXECUTION_ID, "
 			+ "START_TIME, END_TIME, TASK_NAME, EXIT_CODE, "
 			+ "EXIT_MESSAGE, LAST_UPDATED, STATUS_CODE "
+			+ "from %PREFIX%EXECUTION order by START_TIME desc, TASK_EXECUTION_ID desc "
+			+ "LIMIT ? OFFSET ?";
```
mminella: Have you tested this against Oracle?  I didn't think they supported `LIMIT` `OFFEST` syntax.

mminella:
LGTM.  Merged as 0e90778f20b7c942319bce9da3ae0979661f54a0.

Issue #39, opened 2015-12-11
User wants EnableTask at SpringApplication level vs. *Runner


cppwfs:
Is this possible or do we stay at the *Runner Level

PR #45, opened 2015-12-21
Refactor of Task API

mminella:
This change removes the Spring Cloud Task dependency upon a
CommandLineRunner from boot and moves the handling of the task lifecycle
closer to the "edge" of a Spring Boot application.

With this change, now a developer simply adds @EnableTask to their
configuration somewhere and the task lifecycle will be recorded.

Resolves spring-cloud/spring-cloud-task#39


pom.xml
```diff
@@ -8,7 +8,7 @@
 				<!-- Import dependency management from Spring Boot -->
 				<groupId>org.springframework.boot</groupId>
 				<artifactId>spring-boot-dependencies</artifactId>
-				<version>1.3.0.RELEASE</version>
+				<version>1.3.1.RELEASE</version>
```
cppwfs: Do we want to move to snapshot instead of release?

cppwfs: Need to update the README.adoc

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -0,0 +1,162 @@
...
+		else if(applicationEvent instanceof ApplicationFailedEvent) {
+			doTaskFailed(((ApplicationFailedEvent) applicationEvent).getException());
+		}
+	}
+
+	private void doTaskFailed(Throwable exception) {
+
+		if(started) {
+			this.taskExecution.setEndTime(new Date());
+			//TODO: get exit code from new event thrown by Boot.
```
cppwfs: Not sure that this will be valid if there is an uncaught exception.  i.e. the status of the boot app could be zero even though the app technically failed.  The suggestion would be that if the exit code is 0 "as returned from boot"  we should still record a 1.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -0,0 +1,162 @@
...
+			this.taskExecution.setExitCode(1);
+			this.taskExecution.setExitMessage(stackTraceToString(exception));
+
+			taskRepository.update(taskExecution);
+		}
+		else  {
+			logger.warn("An event to fail a task has been received for a task that has " +
```
cppwfs: log an error possibly.  Thoughts?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -0,0 +1,162 @@
...
+	private void doTaskEnd() {
+		if(started) {
+			this.taskExecution.setEndTime(new Date());
+			//TODO: get exit code from new event thrown by Boot.
+			this.taskExecution.setExitCode(0);
+
+			taskRepository.update(taskExecution);
+		}
+		else {
+			logger.warn("An event to end a task has been received for a task that has " +
```
cppwfs: log an error possibly? Thoughts?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	private void doTaskStart() {
+
+		if(!started) {
+
+			String executionId = UUID.randomUUID().toString();
+			this.taskExecution = new TaskExecution();
+
+			this.taskExecution.setTaskName(applicationContext.getId());
```
cppwfs: We should use the spring.application.name boot property instead of the applicationContext.getId.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -0,0 +1,162 @@
...
+
+	/**
+	 * @param taskRepository The repository to record executions in.
+	 */
+	public TaskLifecycleListener(TaskRepository taskRepository) {
+		this.taskRepository = taskRepository;
+	}
+
+	/**
+	 * @param applicationEvent The application being listened for.
```
cppwfs: add a description as to what events this method is going to handle.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/listener/TaskLifecycleListenerTests.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+public class TaskLifecycleListenerTests {
+
+	@Autowired
+	private TaskLifecycleListener listener;
+
+	@Test
+	public void testTaskCreate() {
+		AnnotationConfigApplicationContext createContext = new AnnotationConfigApplicationContext();
+		createContext.register(TestDefaultConfiguration.class, PropertyPlaceholderAutoConfiguration.class);
```
cppwfs: We can create a `@Before` method to handle the initializtion of the context and the registration of the configuration classes (i.e. lines 48-49, 63-64, 78-79).  Then a `@After` method to close the context.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/listener/TaskLifecycleListenerTests.java
```diff
@@ -0,0 +1,91 @@
...
+		createContext.refresh();
+		final Appender mockAppender = TestVerifierUtils.getMockAppender();
+		this.listener = createContext.getBean(TaskLifecycleListener.class);
+		listener.onApplicationEvent(new ApplicationFailedEvent(new SpringApplication(), new String[]{}, createContext, new RuntimeException("This was expected")));
+		TestVerifierUtils.verifyLogEntryExists(mockAppender,
+				"Updating: TaskExecution{executionId='" +
+						listener.getTaskExecution().getExecutionId());
+		assertEquals("Update should report that exit code is zero",
```
cppwfs: Should read `Update should report that exit code is one`

cppwfs:
Need to update the README.adoc to update the sample. i.e. we're no long using the `@Task `annotation.

cppwfs:
Need to update the timestamp sample in the spring-cloud-task-samples module.  i.e. removing and `@Task` and changing the import for `@EnableTask`

pom.xml
mminella: Agreed.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
mminella: That means we're overriding what the app has requested and not a good idea IMHO.  We're here to record what the app did...not change what the app did.  To your point, if a batch job run via boot throws an uncaught exception, it's eventually caught by the framework and the boot jar completes successfully and I think that's ok.  We can't be held responsible for things we can't control.  Thoughts (may be worth an offline conversation)?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
mminella: Agreed.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
mminella: Agreed.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
mminella: As discussed, this is the expected way to obtain the fully processed value per https://github.com/spring-projects/spring-boot/blob/master/spring-boot/src/main/java/org/springframework/boot/context/ContextIdApplicationContextInitializer.java

mminella:
Updates made including the README and the timestamp task sample.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
cppwfs: Agreed.  Really depends on the implementation that boot will be adding on uncaught exceptions.  This comment is based on the current implementation (which is going away)

cppwfs:
LGTM.  Let's discuss the Exception error code, then I'll merge.

cppwfs:
Squashed, Merged

Issue #47, opened 2015-12-30
Developer wants to use mvnw for task builds


cppwfs:
To remain consistent with other cloud projects SCT needs to build using mvnw

PR #51, opened 2015-12-31
SCT-47 Support Maven Wrapper

cppwfs:
resolves spring-cloud/spring-cloud-task#47


.mvn/maven.config
```diff
@@ -0,0 +1 @@
+-DaltSnapshotDeploymentRepository=repo.spring.io::default::https://repo.spring.io/libs-snapshot-local -s settings.xml -P spring
```
markpollack: Should be .settings.xml

cppwfs: The spring task project the settings.xml file does not have the `.` prefix.

trisberg: it should use .settings.xml to be consistent with other  projects

trisberg:
Can we also add a default profile that includes the repos like here -- https://github.com/spring-cloud/spring-cloud-stream-modules/blob/master/pom.xml#L212

.mvn/maven.config
mminella: -1 for changing the name to `.settings.xml`.  If we are going to require it's use, it shouldn't be hidden.  Either set it up so it's not needed, or don't hide it IMHO.

cppwfs: Will do

mminella:
Squashed and merged as 30497e62584879130c80b2fe9431c864f265c35a.

PR #53, opened 2016-01-06
TaskExecution should use long id

cppwfs:
* Replace String executionId with a long executionId
* Add externalExecutionID that is a String

resolves spring-cloud/spring-cloud-task#34


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -110,7 +112,9 @@
 
 	private DataSource dataSource;
 
-	Map<String, Order> orderMap;
+	private Map<String, Order> orderMap;
+
+	private DataFieldMaxValueIncrementer jobIncrementer;
```
mminella: Doing some copy/paste here?  ;)

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -248,6 +254,14 @@ public long getTaskExecutionCount() {
 		return new PageImpl<TaskExecution>(resultList, pageable, getTaskExecutionCount());
 	}
 
+	public void setJobIncrementer(DataFieldMaxValueIncrementer jobIncrementer) {
+		this.jobIncrementer = jobIncrementer;
+	}
```
mminella: Same copy/paste issue

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
cppwfs: DOH!

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-hsqldb.sql
```diff
@@ -11,8 +12,12 @@ CREATE TABLE TASK_EXECUTION  (
...
 CREATE TABLE TASK_EXECUTION_PARAMS  (
-	TASK_EXECUTION_ID VARCHAR(100) NOT NULL ,
+	TASK_EXECUTION_ID BIGINT NOT NULL ,
 	TASK_PARAM VARCHAR(250) ,
 	constraint TASK_EXEC_PARAMS_FK foreign key (TASK_EXECUTION_ID)
 	references TASK_EXECUTION(TASK_EXECUTION_ID)
 ) ;
+
+CREATE TABLE TASK_SEQ (
+	ID BIGINT IDENTITY
```
mminella: No unique key for HSQL?

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-mysql.sql
```diff
@@ -11,8 +12,17 @@ CREATE TABLE TASK_EXECUTION  (
...
 ) ;
+
+
+CREATE TABLE TASK_SEQ (
+	ID BIGINT NOT NULL,
+	UNIQUE_KEY CHAR(1) NOT NULL,
+	constraint UNIQUE_KEY_UN unique (UNIQUE_KEY)
+) ENGINE=InnoDB;
+
+INSERT INTO TASK_SEQ (ID, UNIQUE_KEY) select * from (select 0 as ID, '0' as UNIQUE_KEY) as tmp where not exists(select * from TASK_SEQ);
```
mminella: Do you need the where not exists clause given the unique key?

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/util/TestDBUtils.java
```diff
@@ -55,7 +60,7 @@
 	 * @return taskExecution
 	 */
 	public static TaskExecution getTaskExecutionFromDB(DataSource dataSource,
-													   String taskExecutionId) {
+													   long taskExecutionId) {
```
mminella: Odd formatting...

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-hsqldb.sql
cppwfs: Don't think we need this because the id is  a identity column.

mminella: If I run the population script once...run a few tasks...then run the script again, won't you get two rows?

cppwfs: Not at this point.  The script will attempt to run, but fail when it tries to create task_execution, thus the entry is not re-added.

cppwfs: Not at this point.  The script will attempt to run, but fail when it tries to create task_execution, thus the entry is not re-added.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/TaskExecution.java
```diff
@@ -90,11 +96,11 @@ public TaskExecution(String executionId, int exitCode, String taskName,
 		setEndTime(endTime);
 	}
 
-	public String getExecutionId() {
+	public long getExecutionId() {
 		return executionId;
 	}
 
-	public void setExecutionId(String executionId) {
+	public void setExecutionId(long executionId) {
```
mminella: Hmmm...Should we allow this to be set?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -52,15 +53,16 @@
 public class JdbcTaskExecutionDao implements TaskExecutionDao {
 
 
-	public static String SELECT_CLAUSE = "TASK_EXECUTION_ID, "
+	public static String SELECT_CLAUSE = "TASK_EXECUTION_ID, TASK_EXTERNAL_EXECUTION_ID, "
```
mminella: Being nit picky here but we probably don't need to prefix the name of the column with task...

mminella:
Tweaked per disccusion.  Squashed and merged as 584cfb90f5c4c66279bbfb901d7e5dc7dae68f37

PR #64, opened 2016-01-19
Adding NOCACHE to sequence TASK_SEQ

### LINKED ISSUE #63: Use NOCACHE for Oracle sequence
When creating the Oracle sequence TASK_SEQ we should use NOCACHE to avoid the default to CACHE 20. 

With the current default, every time we run a task, we increment the sequence by 20 while only using one of these sequence values.

trisberg:
Resolves #63


cppwfs:
LGTM Tested fine on Oracle

cppwfs:
Merged.

Issue #69, opened 2016-01-21
Users need sequence table to use MYISAM


cppwfs:
As discussed here the engine for the table should be ISAM vs. InnoDB.

http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/jdbc/support/incrementer/MySQLMaxValueIncrementer.html

trisberg:
Just to clarify, it should be MYISAM for the sequence table and the other tables should be InnoDB

PR #70, opened 2016-01-22
SCT-69 Mysql script uses correct engines for each tables.

cppwfs:
resolves spring-cloud/spring-cloud-task#69


trisberg:
Merged via commit eb995ee0376d9aa936578a33d412d25be33ce1d9

Issue #48, opened 2015-12-30
User wants boot exit code recorded when boot app ends


cppwfs:
When Spring boot supports exit events with exit code, SCT will need to use that event instead of ContextClosedEvent.

PR #71, opened 2016-01-22
Use spring Boot events to get exit code

mminella:
Spring Boot now provides the result of an ExitCodeExceptionMapper via an
event called ExitCodeEvent.  This commit takes advantage of this new
event.

Resolves spring-cloud/spring-cloud-task#48


cppwfs:
LGTM

Issue #72, opened 2016-01-25
Developers want to remove unused taskExecution Fields


cppwfs:
Remove externalExecutionID & StatusCode from classes and schema.

PR #73, opened 2016-01-25
SCT-72 Remove externalExecutionID & StatusCode

cppwfs:
Will add these back when and if they are needed.

resolves spring-cloud/spring-cloud-task#72


trisberg:
Merged via commit 2efa75ff16ee17141febd7b4304faa1700f0cbb9

Issue #58, opened 2016-01-14
Create a BOM for users to add SCT dependencies


mminella:
As a user, I want to be able to use Spring Cloud Task (SCT) and manage the version of Spring Boot independently.

Back Story: Since SCT depends on Boot, we need a BOM that contains only the dependencies for SCT so that users can layer that on top of the Boot dependencies they are already bringing in.

PR #75, opened 2016-01-25
SCT-58 Add bom to SCT

cppwfs:
Resolves spring-cloud/spring-cloud-task#58

### DIFF
--- a/spring-cloud-task-dependencies/pom.xml
+++ b/spring-cloud-task-dependencies/pom.xml
@@ -1,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<groupId>org.springframework.cloud</groupId>
+	<modelVersion>4.0.0</modelVersion>
+	<artifactId>spring-cloud-task-dependencies</artifactId>
+	<version>1.0.0.BUILD-SNAPSHOT</version>
+	<packaging>pom</packaging>
+	<name>spring-cloud-task-dependencies</name>
+	<description>Spring Cloud Task Dependencies</description>
+	<properties>
+		<spring-boot.version>1.3.2.RELEASE</spring-boot.version>
+	</properties>
+
+	<parent>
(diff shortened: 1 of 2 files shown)


spring-cloud-task-dependencies/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+	<name>spring-cloud-task-dependencies</name>
+	<description>Spring Cloud Task Dependencies</description>
+	<properties>
+		<spring-boot.version>1.3.2.RELEASE</spring-boot.version>
+	</properties>
+
+	<parent>
+		<groupId>org.springframework.boot</groupId>
+		<artifactId>spring-boot-starter-parent</artifactId>
+		<version>1.3.2.RELEASE</version>
```
mminella: I wouldn't expect this have a parent at all...

spring-cloud-task-dependencies/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+	</parent>
+
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-task-core</artifactId>
+				<version>${project.version}</version>
+			</dependency>
```
mminella: We need to exclude the boot dependencies here, don't we?

spring-cloud-task-dependencies/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-task-parent</artifactId>
+				<version>${project.version}</version>
+			</dependency>
```
mminella: Doesn't this create a circular dependency?

spring-cloud-task-dependencies/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+			<dependency>
+				<groupId>org.springframework.boot</groupId>
+				<artifactId>spring-boot-dependencies</artifactId>
+				<version>${spring-boot.version}</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
```
mminella: How is this being used?

mminella:
I'd expect that this BOM consist of all of the Spring Cloud Task bits minus any Boot bits.  Is that incorrect?

mminella:
Squashed and merged as bed52091a9b0f98bd7f095214a62254ad7cc052d

Issue #85, opened 2016-02-10
Align maven build conventions with those of other spring cloud projects


markpollack:
The delta seems to be that the pom.xml should contain the spring profile that lists the spring release/milestone/snapshot repositories and that the current settings.xml file should be renamed to .settings.xml as it is for use w/ some IDE configurations and not related to general command line building.

PR #86, opened 2016-02-10
GH-85 Align maven build to spring cloud conventions

markpollack:
* move settings.xml to .settings.xml
* add spring profile to pom.xml with spring repository locations


cppwfs:
Let's Discuss this with Michael.  We have visited this topic 2 times already.  i.e. the name of the .settings.xml vs settings.xml  and the profiles.   
Just want to make sure we make a decision and stick to it :-)

markpollack:
Had a chat, seems on board.

cppwfs:
Sounds great!

Issue #84, opened 2016-02-10
SimpleTaskConfiguration needs to initialize db as a bean


cppwfs:
Using Boot's DataSourceInitializer as a pattern.  i.e. Create a askSourceInitializer bean.

Also rename spring.class.initialize.enable to spring.task.initialize.enable

PR #87, opened 2016-02-11
Refactored database initialization

mminella:
This commit introduces the `TaskRepositoryInitializer` component.  This
component will initialize the `DataSource` when appropriate and silently do
nothing otherwise.  It can be customized to initialize a specific
`DataSource` as required.

Resolves spring-cloud/spring-cloud-task#84

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
```diff
@@ -124,12 +128,7 @@ private TaskConfigurer getDefaultConfigurer(Collection<TaskConfigurer> configure
 				return this.configurer;
 			}
 			else {
-				DataSource dataSource = dataSources.iterator().next();
-				if(taskInitializationEnable) {
-					logger.debug("Initializing Task Schema");
-					TaskDatabaseInitializer.initializeDatabase(dataSource, resourceLoader);
-				}
-				this.configurer = new DefaultTaskConfigurer(dataSource);
+				this.configurer = new DefaultTaskConfigurer(this.dataSources.iterator().next());
```
cppwfs: For line 123 we probably want to add dataSources.size() == 1

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/configuration/TestConfiguration.java
```diff
@@ -47,6 +47,16 @@
 	private ResourceLoader resourceLoader;
 
 	@Bean
+	public TaskRepositoryInitializer taskDatabaseInitializer() throws Exception {
```
cppwfs: Rename to taskRepositoryInitializer

cppwfs:
LGTM

PR #90, opened 2016-02-22
SCT-82 Exit Code needs to be null while task is running

cppwfs:
TaskLifecycleListener will set a TaskExecution's exitCode to null upon taskStart in the repository.
Only at the end of the execution will it store the exitCode.


cppwfs:
Closing to as to associate it with SCT-82 story.

Issue #79, opened 2016-01-28
Add MS SQL Server as supported databases


trisberg:
We should support the most commonly used databases, so this would mean we should add MS SQL Server his would also be good for cloud deployments since Azure has SQL Server 

For popularity numbers see - http://db-engines.com/en/ranking_osvsc

cppwfs:
Split this to 2 stories.  Story #89 covers DB2

PR #92, opened 2016-02-22
SCT-79 Support sqlserver

cppwfs:
resolves spring-cloud/spring-cloud-task#79


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/database/support/OraclePagingQueryProvider.java
```diff
@@ -29,7 +29,22 @@
 	@Override
 	public String getPageQuery(Pageable pageable) {
 		int offset = pageable.getOffset()+1;
-		return SqlPagingQueryUtils.generateRowNumSqlQueryWithNesting(this, getSelectClause(), getSelectClause(), false, "TMP_ROW_NUM >= "
+		return generateRowNumSqlQueryWithNesting(getSelectClause(), getSelectClause(), false, "TMP_ROW_NUM >= "
 				+ offset + " AND TMP_ROW_NUM < " + (offset+pageable.getPageSize()));
 	}
+
+	private String generateRowNumSqlQueryWithNesting(String innerSelectClause,
```
mminella: Since this is a private method and you pass in the same for both inner and outer, why not just pass it in once?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/database/support/SqlServerPagingQueryProvider.java
```diff
@@ -0,0 +1,51 @@
...
+public class SqlServerPagingQueryProvider extends AbstractSqlPagingQueryProvider{
+
+	@Override
+	public String getPageQuery(Pageable pageable) {
+		int offset = pageable.getOffset()+1;
+		return generateRowNumSqlQueryWithNesting(getSelectClause(), getSelectClause(), false, "TMP_ROW_NUM >= "
+				+ offset + " AND TMP_ROW_NUM < " + (offset+pageable.getPageSize()));
+	}
+
+	private String generateRowNumSqlQueryWithNesting(String innerSelectClause,
```
mminella: Same comment from before about the duplicate inner/outer param

spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/schema-sqlserver.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE TASK_EXECUTION  (
+  TASK_EXECUTION_ID BIGINT  NOT NULL PRIMARY KEY ,
+  START_TIME DATETIME DEFAULT NULL ,
+  END_TIME DATETIME DEFAULT NULL ,
+  TASK_NAME  VARCHAR(100) ,
+  EXIT_CODE INTEGER ,
+  EXIT_MESSAGE VARCHAR(2500) ,
```
mminella: Given that we store stack traces here...do we want to convert these to something larger (for example `TEXT` on MySql)?

cppwfs: Will use varchar(MAX) because TEXT is being deprecated.    http://stackoverflow.com/questions/564755/sql-server-text-type-vs-varchar-data-type

mminella: This was meant as a more general statement (across all options), not just specifically this impl.

mminella:
Squashed and merged as 91fa6be0926150fb91c6b81d18b0d80ea42583c2

Issue #35, opened 2015-12-11
TaskExecutionListener


mminella:
As a developer, I want to be able to execute code before and/or after the execution of a task.

mminella:
This may be a boot thing (since it's essentially wrapping the execution of a CommandLineRunner).

PR #94, opened 2016-02-23
SCT-35 Tasks Support TaskExecutionListener

cppwfs:
resolves spring-cloud/spring-cloud-task#35

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskExecutionListener.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskExecutionListener.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.task.listener;
+
+import org.springframework.cloud.task.repository.TaskExecution;
+import org.springframework.cloud.task.repository.TaskRepository;
+
+/**
+ * The listener interface for receiving task execution events.
+ * @author Glenn Renfro
+ */
+public interface TaskExecutionListener {
+
+	/**
+	 * Invoked after the {@link TaskExecution} has been stored in the {@link TaskRepository}.
+	 * @param taskExecution instance containing the information about the current task.
+	 */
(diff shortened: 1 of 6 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -169,5 +178,30 @@ private void doTaskStart() {
...
+			}
+		}
+	}
+
+	private void invokeOnTaskError(TaskExecution taskExecution){
+		if (taskExecutionListeners != null) {
+			for (TaskExecutionListener taskExecutionListener : taskExecutionListeners) {
+				taskExecutionListener.onTaskFailed(taskExecution);
+			}
+		}
```
mminella: For all of these, we need to send a copy of the `TaskExecution` to the listener and limit what we copy back.  Otherwise the user could manipulate anything to detrimental consequences.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskExecutionListener.java
```diff
@@ -0,0 +1,49 @@
...
+	 */
+	public void onTaskEnd(TaskExecution taskExecution);
+
+	/**
+	 * Invoked if an uncaught exception occurs during a task execution.  This invocation
+	 * will occur before the {@link TaskExecution} has been updated in the {@link TaskRepository}
+	 * and before the onTaskEnd is called.
+	 * @param taskExecution instance containing the information about the current task.
+	 */
+	public void onTaskFailed(TaskExecution taskExecution);
```
mminella: I thought we agreed that the exception would be passed here as well?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -169,5 +179,42 @@ private void doTaskStart() {
...
+	}
+
+	private void invokeOnTaskError(TaskExecution taskExecution, Throwable throwable){
+		if (taskExecutionListeners != null) {
+			for (TaskExecutionListener taskExecutionListener : taskExecutionListeners) {
+				taskExecutionListener.onTaskFailed(getTaskExecutionCopy(taskExecution),
+						throwable);
+			}
+		}
+	}
```
mminella: So it's good that we're providing a copy of the `TaskExecution`.  I'm wondering if we should allow anything to be copied back (edit message is one thing I'm wondering)...

cppwfs: Maybe a String return val?

mminella: That locks us in from a contract perspective.  What about just declaring in documentation what we plan on copying back over?

cppwfs: Agreed, I'll leave the interface alone, but not sure I understand what we add to the docs.

mminella:
LGTM.  Squashed and merged as 20a56c9581349d377e0a22acb0f8c0f2a86c24a7

Issue #95, opened 2016-02-23
Developer needs to catch TaskExecution events via annotations


cppwfs:
Developer should be able to associate the following annotations to methods so that they can be invoked when the event occurs in the task.

@BeforeTask`, `@AfterTask`, `@AfterTaskFails`

PR #98, opened 2016-02-29
SCT-95 Support TaskExecutionListener Annotations

cppwfs:
resolves spring-cloud/spring-cloud-task#95


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/AfterTask.java
```diff
@@ -0,0 +1,44 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * <p>
+ * Enable a method to be invoked after a task has ended but prior to TaskRepository update.
+ *
+ * <pre class="code">
+ * public class MyListener {
+
+ * &#064;AfterTask
```
mminella: Does that format correctly (looks like the `@AfterTask` wouldn't be indented in the javadoc).

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/BeforeTask.java
```diff
@@ -0,0 +1,44 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * <p>
+ * Enable a method to be invoked after repository has been updated but before task has started.
+ *
+ * <pre class="code">
+ * public class MyListener {
+
+ * &#064;BeforeTask
```
mminella: Same formatting question

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/AfterTask.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * <p>
+ * Enable a method to be invoked after a task has ended but prior to TaskRepository update.
```
mminella: Instead of re-stating the contract here, do we want to just reference the `TaskExecutionListener#afterTask` method saying it obeys the same contract?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/BeforeTask.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * <p>
+ * Enable a method to be invoked after repository has been updated but before task has started.
```
mminella: Instead of re-stating the contract here, do we want to just reference the `TaskExecutionListener#beforeTask` method saying it obeys the same contract?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/AfterTask.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * <pre class="code">
+ * public class MyListener {
+
+ * &#064;AfterTask
+ * 	public void  doSomething(TaskExecution taskExecution) {
+ * 	}
+ * }
+ * </pre>
```
mminella: Missing closing `</p>`.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/BeforeTask.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * <pre class="code">
+ * public class MyListener {
+
+ * &#064;BeforeTask
+ * 	public void  doSomething(TaskExecution taskExecution) {
+ * 	}
+ * }
+ * </pre>
```
mminella: Missing closing `</p>`.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/FailedTask.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * <pre class="code">
+ * public class MyListener {
+
+ * &#064;FailedTask
+ * 	public void  doSomething(TaskExecution taskExecution, Throwable throwable) {
+ * 	}
+ * }
+ * </pre>
+ *
```
mminella: Same three comments on javadoc here (formatting the annotation, contract reference and missing `</p>`).

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,243 @@
...
+ */
+public class TaskListenerExecutor {
+
+	private ConfigurableApplicationContext context;
+
+	private final static Logger logger = LoggerFactory.getLogger(TaskListenerExecutor.class);
+
+
+	private final Set<Class<?>> nonAnnotatedClasses =
+			Collections.newSetFromMap(new ConcurrentHashMap<Class<?>, Boolean>(64));
```
mminella: Why 64?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,243 @@
...
+import org.springframework.core.annotation.AnnotationUtils;
+
+/**
+ * Identifies all beans that contain a TaskExecutionListener annotation and stores the
+ * associated method so that it can be called by the {@link TaskExecutionListener} at the
+ * appropriate time.
+ *
+ * @author Glenn Renfro
+ */
+public class TaskListenerExecutor {
```
mminella: Why not just have this implement `TaskExecutionListener` so it's treated like any other?  That eliminates the special calling code in the `TaskLifecycleListener`...

cppwfs: good catch!

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+	private void executeTaskListener(TaskExecution taskExecution, Set<Method> methods, Map<Method, Object> instances){
+		for (Method method : methods) {
+			try {
+				method.invoke(instances.get(method),taskExecution);
+			}
+			catch (IllegalAccessException e) {
+				throw new IllegalStateException("@BeforeTask and @AfterTask annotated methods must be public.");
+			}
+			catch (InvocationTargetException e) {
+				throw new IllegalStateException("Failed to process @BeforeTask or @AfterTask" +
```
mminella: Does an `IllegalStateException` make sense here?  Is it really a state issue?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+	private void initializeExecutor( ) {
+		ConfigurableListableBeanFactory factory = context.getBeanFactory();
+		for( String beanName : context.getBeanNamesForType(Object.class)) {
+
+			if (!ScopedProxyUtils.isScopedTarget(beanName)) {
+				Class<?> type = null;
+				try {
+					type = AutoProxyUtils.determineTargetClass(this.context.getBeanFactory(), beanName);
+				}
+				catch (Throwable ex) {
```
mminella: Swallowing `Throwable`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+						logger.debug("Could not resolve target class for bean with name '" + beanName + "'", ex);
+					}
+				}
+				if (type != null) {
+					if (ScopedObject.class.isAssignableFrom(type)) {
+						try {
+							type = AutoProxyUtils.determineTargetClass(this.context.getBeanFactory(),
+									ScopedProxyUtils.getTargetBeanName(beanName));
+						}
+						catch (Throwable ex) {
```
mminella: Again, swallowing `Throwable`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+							// An invalid scoped proxy arrangement - let's ignore it.
+							if (logger.isDebugEnabled()) {
+								logger.debug("Could not resolve target bean for scoped proxy '" + beanName + "'", ex);
+							}
+						}
+					}
+					try {
+						processBean(beanName, type);
+					}
+					catch (Throwable ex) {
```
mminella: Again, swallowing `Throwable`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+			catch (IllegalArgumentException e){
+				throw new IllegalStateException("taskExecution and throwable parameters "
+						+ "are required for @FailedTask annotated methods");
+			}
+		}
+	}
+
+	private void initializeExecutor( ) {
+		ConfigurableListableBeanFactory factory = context.getBeanFactory();
+		for( String beanName : context.getBeanNamesForType(Object.class)) {
```
mminella: Can't you use `context.getBeanDefinitionNames()`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+			Collections.newSetFromMap(new ConcurrentHashMap<Class<?>, Boolean>());
+
+	private Map<Method, BeforeTask> beforeTaskMethods;
+	private Map<Method, Object> beforeTaskInstances;
+
+	private Map<Method, AfterTask> afterTaskMethods;
+	private Map<Method, Object> afterTaskInstances;
+
+	private Map<Method, FailedTask> failedTaskMethods;
+	private Map<Method, Object> failedTaskInstances;
```
mminella: I may not be following here, but why do you need to store a `Map<Method, ANNOTATION>` and `Map<Method, INSTANCE>`?  Since you have a map for each annotation type, why isn't the latter sufficient?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	private Map<Method, FailedTask> getFailedTaskMethods(final Class<?> type){
+		return MethodIntrospector.selectMethods(type,
+				new MethodIntrospector.MetadataLookup<FailedTask>() {
+					@Override
+					public FailedTask inspect(Method method) {
+						return AnnotationUtils.findAnnotation(method, FailedTask.class);
+					}
+				});
```
mminella: If you do need these three methods (see previous comment about the `Map<Method, ANNOTATION>`, can't you do this once with generics?  Something like (I haven't tried this so feel free to call BS):
```
	public static class MethodGetter<T extends Annotation> {
		public Map<Method, T> getMethods(final Class<?> type, final Class<T> annotationClass){
			return MethodIntrospector.selectMethods(type,
					new MethodIntrospector.MetadataLookup<T>() {
						@Override
						public T inspect(Method method) {
							return AnnotationUtils.findAnnotation(method, annotationClass);
						}
					});
		}
	}
...

Map<Method, BeforeTask> beforeTask = new MethodGetter<BeforeTask>().getMethods(type, BeforeTask.class);
Map<Method, AfterTask> afterTask = new MethodGetter<AfterTask>().getMethods(type, AfterTask.class);
Map<Method, FailedTask> failedTask = new MethodGetter<FailedTask>().getMethods(type, FailedTask.class);
```

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+						"annotation because: ", e);
+			}
+			catch (IllegalArgumentException e){
+				throw new IllegalStateException("taskExecution and throwable parameters "
+						+ "are required for @FailedTask annotated methods");
+			}
+		}
+	}
+
+	private void initializeExecutor( ) {
```
mminella: Wondering if this code belongs in a factory...

spring-cloud-task-docs/src/main/asciidoc/features.adoc
```diff
@@ -185,3 +185,30 @@ notified for the following events:
 marking the final state of the task.
 . `onTaskFailed` - prior to the `onTaskEnd` method being invoked when an unhandled
  exception is thrown by the task.
+
+Spring Cloud Task also allows a user add TaskExecution Listeners to methods within a bean
```
mminella: TaskExecution should have ticks around it.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,246 @@
...
+	@Override
+	public void onTaskFailed(TaskExecution taskExecution, Throwable throwable) {
+		executeTaskListenerWithThrowable(taskExecution, throwable,
+				failedTaskMethods.keySet(),failedTaskInstances);
+	}
+
+	private void executeTaskListener(TaskExecution taskExecution, Set<Method> methods, Map<Method, Object> instances){
+		for (Method method : methods) {
+			try {
+				method.invoke(instances.get(method),taskExecution);
```
mminella: Does this require that the method signature accept the `TaskExecution` parameter?  Is it possible to make the parameter optional on the annotation?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
cppwfs: In this case when determining the target class a lazy bean could thrown an exception.  The beans that utilize the annotations should not be lazy.  I can kick the message to info or warning.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
cppwfs: Similar as above when determining target class, if a bean has a invalid scoped proxy arrangement, it should not cause the task to fail.  we can change the log to a warning.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
mminella: Do we know what the exception is and if so, can we be more targeted?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
cppwfs: Yes but why?

mminella: Flexibility...it's a nice to have.  Not a big deal

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
cppwfs: Yes...  It was spike code that should have been removed.  Doh!

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
cppwfs: hmm let's talk about that.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskListenerExecutor.java
```diff
@@ -0,0 +1,225 @@
...
+	private final Set<Class<?>> nonAnnotatedClasses =
+			Collections.newSetFromMap(new ConcurrentHashMap<Class<?>, Boolean>());
+
+	private Map<Method, Object> beforeTaskInstances;
+
+	private Map<Method, Object> afterTaskInstances;
+
+	private Map<Method, Object> failedTaskInstances;
+
+	private TaskListenerExecutor(){
```
mminella: What is this for?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/annotation/TaskAnnotationException.java
```diff
@@ -0,0 +1,34 @@
...
+
+package org.springframework.cloud.task.listener.annotation;
+
+/**
+ * Is thrown when failure occurs while trying to invoke a method that has been annotated
+ * by a task execution listener.
+ *
+ * @author Glenn Renfro.
+ */
+public class TaskAnnotationException extends RuntimeException {
```
markpollack: The naming here seems to be a bit strange, to have the word 'annotation' in it.  In looking at the high level flow of SI when calling a service activator, in `MethodInvokingMessageHandler` or `MessageHandlingException` at runtime there is either a `MessagingException` or `MessageHandlingException` thrown at runtime. Maybe a similar scheme, `TaskExecutionException` that is a subclass of `TaskException`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -62,6 +63,9 @@
 	@Autowired(required = false)
 	Collection<TaskExecutionListener> taskExecutionListeners;
 
+	@Autowired
+	private ConfigurableApplicationContext context;
+
```
mminella: This was added but never used.

mminella:
Updated, squashed, and merged as f4860500863ff0aa795cc8ed9c7e0054d8947f42

Issue #46, opened 2015-12-30
Add ability to associate batch job with task it was executed within


mminella:
As a developer, when I run a Spring Batch job as a Spring Cloud Task, I want to be able to associate the two.

PR #103, opened 2016-03-09
Add listener to persist relationship between job and task

mminella:
This introduces a listener that stores the association between a Spring
Batch job and the task it was executed within.

Resolves spring-cloud/spring-cloud-task#46


spring-cloud-task-samples/batch-job/src/test/java/demo/BatchJobApplicationTests.java
```diff
@@ -0,0 +1,80 @@
...
+
+		assertEquals(0, SpringApplication.exit(SpringApplication
+				.run(BatchJobApplication.class)));
+
+		String output = this.outputCapture.toString();
+		assertTrue("Unable to find the timestamp: " + output,
+				output.contains(JOB_RUN_MESSAGE));
+		assertTrue("Test results do not show create task message: " + output,
+				output.contains(CREATE_TASK_MESSAGE));
+		assertTrue("Test results do not show success message: " + output,
```
cppwfs: Have to update tests to support that Update log message is only debug now.  Will do this on merge.

spring-cloud-task-batch/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+	<description>Module for use when combining Spring Cloud Task with Spring Batch</description>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-task-core</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
```
cppwfs: This should have a scope of test.  Will fix on merge.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/listener/support/JdbcTaskBatchDao.java
```diff
@@ -0,0 +1,73 @@
...
+		jdbcTemplate.update(getQuery(INSERT_STATEMENT), taskExecution.getExecutionId(), jobExecution.getId());
+	}
+
+	/**
+	 * The table prefix for the task batch table.
+	 *
+	 * @param tablePrefix defaults to {@link JdbcTaskExecutionDao#DEFAULT_TABLE_PREFIX}.
+	 */
+	public void setTablePrefix(String tablePrefix) {
+		Assert.notNull(tablePrefix, "Null is not allowed as a tablePrfix (use an empty string if you don't want a prefix at all).");
```
cppwfs: Prefix is misspelled will fix on merge.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchExecutionListenerFactoryBean.java
```diff
@@ -0,0 +1,119 @@
...
+ * listener if there is a {@link DataSource} available.  Otherwise, builds a listener that
+ * uses the map based implementation.
+ *
+ * @author Michael Minella
+ */
+public class TaskBatchExecutionListenerFactoryBean implements FactoryBean<TaskBatchExecutionListener> {
+
+	private ConfigurableApplicationContext context;
+
+	private TaskBatchExecutionListener listener;
```
cppwfs: We can move this to local scope.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchExecutionListenerFactoryBean.java
```diff
@@ -0,0 +1,119 @@
...
+	public boolean isSingleton() {
+		return true;
+	}
+
+	public void setDataSourceName(String dataSourceName) {
+		this.dataSourceName = dataSourceName;
+	}
+
+	private MapTaskBatchDao getMapTaskBatchDao() throws Exception {
+		Field taskExecutionDaoField = ReflectionUtils.findField(SimpleTaskExplorer.class, "taskExecutionDao");
```
cppwfs: What if the user is not using SimpleTaskExplorer?

cppwfs: Nevermind.  Its MapTaskExecutionDao, they should only be using simple.  (Answered my own question)

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/listener/TaskBatchExecutionListener.java
```diff
@@ -0,0 +1,77 @@
...
+
+		this.taskBatchDao = taskBatchDao;
+	}
+
+	@Override
+	public void onTaskStartup(TaskExecution taskExecution) {
+		this.taskExecution = taskExecution;
+	}
+
+	@Override
```
cppwfs: We can use the Annotations now and remove the //no-op

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/listener/support/MapTaskBatchDao.java
```diff
@@ -0,0 +1,56 @@
...
+ * purposes only!</note>
+ *
+ * @author Michael Minella
+ */
+public class MapTaskBatchDao implements TaskBatchDao {
+
+	private Map<Long, Set<Long>> relationships;
+
+	public MapTaskBatchDao(Map<Long, Set<Long>> relationships) {
+		this.relationships = relationships;
```
cppwfs: assert not null on relationships.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -241,6 +249,43 @@ public long getNextExecutionId(){
...
+							while(resultSet.next()) {
+								jobExecutionIds.add(resultSet.getLong("JOB_EXECUTION_ID"));
+							}
+
+							return jobExecutionIds;
+						}
+					});
+		}
+		catch (DataAccessException e) {
+			return new HashSet<>(0);
```
cppwfs: Why not use Collections.emptySet()

mminella: Because I'm stupid?  ;)

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -59,7 +62,12 @@ public void updateTaskExecution(TaskExecution taskExecution) {
 
 	@Override
 	public TaskExecution getTaskExecution(long executionId) {
-		return taskExecutions.get(executionId);
+		if(taskExecutions.containsKey(executionId)) {
```
cppwfs: Why check for contains key.  If the map does not have the executionId it should return null.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchExecutionListenerFactoryBean.java
mminella: Actually, the intent was to cache the instance so that multiple calls to `getObject` would return the same instance.

spring-cloud-task-batch/pom.xml
mminella: Both of these dependencies should be that way.  Good catch.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/TaskExecutionDao.java
```diff
@@ -105,11 +106,27 @@
...
+	 * Returns the id of the TaskExecution that the requested Spring Batch job execution
+	 * was executed within the context of.  Returns null if non were found.
+	 *
+	 * @param jobExecutionId the id of the {@link org.springframework.batch.core.JobExecution}
+	 * @return the id of the {@link TaskExecution}
+	 */
+	Long getTaskExecutionIdByJobExecutionId(long jobExecutionId);
+
+	/**
+	 * Returns the
```
cppwfs: Returns the ?  ring from mordor :-D

cppwfs:
Fixed comments above, Merged.

PR #106, opened 2016-03-10
SCT-12 Creates the SCSt sink that can launch tasks.

cppwfs:
resolves  spring-cloud/spring-cloud-task#46


spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
```diff
@@ -0,0 +1,73 @@
...
+	@Autowired
+	public TaskLauncher taskLauncher;
+
+	/**
+	 * Launches a task upon the receipt of a valid TaskLaunchRequest.
+	 * @param message is a TaskLaunchRequest containing the information required to launch
+	 * a task.
+	 */
+	@ServiceActivator(inputChannel = Sink.INPUT)
+	public void taskLauncherSink(Message<?> message) {
```
mbogoevici: I would just take advantage of SIs ability to extract the payload and use TaskLaunchRequest directly on the method.

cppwfs:
Closing this PR because the wrong issue number is associated with it.  I'll reopen with a new PR, and it will include the change requested by Marius.

PR #111, opened 2016-03-22
Changes to support putting SCDF back onto snapshot

cppwfs:
* Only configure the BatchListener if the Job.class has a bean instance
* Add the task-batch to the task-dependencies pom.

Resolves spring-cloud/spring-cloud-dataflow#450


spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchAutoConfiguration.java
```diff
@@ -35,7 +35,7 @@
  * @author Michael Minella
  */
 @Configuration
-@ConditionalOnClass({Job.class, EnableTask.class})
+@ConditionalOnBean({Job.class})
```
mminella: While this proved the point...we should also make this conditional on a property allowing a user to turn this functionality off if they want.

PR #118, opened 2016-04-08
Modify poll interval assertion in DeployerPartitionHandlerTests

ilayaperumalg:
- Greater than 20s poll interval check in the assertion might as well needs to be equal to 20s interval as the test could finish before the next second


mminella:
Minor nit pick, it looks like your IDE's ordering of imports doesn't match the Spring standards: https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style .  Statics should be at the end instead of the top.  I can fix on merge.

mminella:
Updated and merged as 9a073631f12be304ca1d73a058da8959e87027ba.  Thanks!

PR #121, opened 2016-04-12
Remove binder dependency from spring-cloud-task-stream

### LINKED ISSUE #120: Remove binder dependency from spring-cloud-task-stream
Currently, `spring-cloud-task-stream` has `redis` binder dependency which need not be required as the binder dependency needs to be provided by the task application itself.

There are also few other dependencies such as `spring-boot-starter-redis` and in test scopes `spring-integration-test`, 'spring-cloud-task-core' which can be removed as well.

ilayaperumalg:
- Spring Cloud Task Stream shouldn't depend on a specific binder implementation; hence removing redis binder from dependencies
   - The binder dependency needs to come from the task application
 - Also removed few other unwanted dependencies

This resolves #120

### DIFF
--- a/spring-cloud-task-stream/pom.xml
+++ b/spring-cloud-task-stream/pom.xml
@@ -6,7 +6,7 @@
 	<artifactId>spring-cloud-task-stream</artifactId>
 	<packaging>jar</packaging>
 	<name>Spring Cloud Task Stream </name>
-	<description>Allows Tasks to be a part of a stream</description>
+	<description>Allows a Task to be a part of a stream</description>
 
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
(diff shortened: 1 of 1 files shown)


spring-cloud-task-stream/pom.xml
```diff
@@ -28,21 +28,6 @@
...
-			<artifactId>spring-boot-starter-redis</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.integration</groupId>
-			<artifactId>spring-integration-test</artifactId>
-			<version>4.2.2.RELEASE</version>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.cloud</groupId>
```
ilayaperumalg: The `spring-cloud-task-core` is needed for the task event listener changes and can be added back when that is added.

cppwfs: In that case should we just leave the spring-cloud-task-core in this PR.  Instead of removing it then re-adding it.  I can do this on merge.

cppwfs:
LGTM Fix on merge.

Issue #122, opened 2016-04-12
Task Processor needs to update default group


cppwfs:
Needs to be io.spring.cloud instead of io.spring

PR #124, opened 2016-04-13
Updates the default group used by the task processor sample

cppwfs:
resolves spring-cloud/spring-cloud-task#122


mminella:
Merged as 1cd18497b8a420f5c450a1590df9c8e289806b6b.

PR #127, opened 2016-04-15
Add integration test for task events

### LINKED ISSUE #126: Add integration test for task event listener
As a developer, I would like to have integration test for Task event listener.

ilayaperumalg:
This resolves #126

### DIFF
--- a/spring-cloud-task-integration-tests/pom.xml
+++ b/spring-cloud-task-integration-tests/pom.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<parent>
+		<artifactId>spring-cloud-task-parent</artifactId>
+		<groupId>org.springframework.cloud</groupId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+	<name>spring-cloud-task-integration-tests</name>
+	<groupId>org.springframework.cloud</groupId>
+	<artifactId>spring-cloud-task-integration-tests</artifactId>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-task-stream</artifactId>
+			<version>1.0.0.BUILD-SNAPSHOT</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-test</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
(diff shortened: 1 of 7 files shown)


spring-cloud-task-stream/pom.xml
```diff
@@ -42,7 +42,13 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-stream-test-support</artifactId>
+			<artifactId>spring-cloud-stream-test-support-internal</artifactId>
+			<version>1.0.0.BUILD-SNAPSHOT</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
```
cppwfs: The TaskEventTests should be skipped if Redis is not available.  At this time they fail, if Redis is not running.

ilayaperumalg: Isn't the `@Rule` on TaskEventTests enough in this case?

cppwfs: When running the tests.  If Redis is up and running locally it works fine.  However if Redis is shutdown and tests are re-run, they fail with `Caused by: redis.clients.jedis.exceptions.JedisConnectionException: java.net.ConnectException: Connection refused`

ilayaperumalg: @cppwfs yeah, sorry it was my mistake to keep the integration test inside along with the unit tests.
With the recent changes to this PR, I have moved the integration test into a separate sub-project (which uses redis binder) and let the unit tests run with the `TestSupportBinder`. This way, the integration test will skip running the test if Redis not available (instead of failing).

spring-cloud-task-integration-tests/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+	<name>spring-cloud-task-integration-tests</name>
+	<groupId>org.springframework.cloud</groupId>
+	<artifactId>spring-cloud-task-integration-tests</artifactId>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-task-stream</artifactId>
+			<version>1.0.0.BUILD-SNAPSHOT</version>
+		</dependency>
```
cppwfs: We can change this to test scope.  Will fix on merge.

spring-cloud-task-integration-tests/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-binder-redis</artifactId>
+			<version>1.0.0.BUILD-SNAPSHOT</version>
+			<scope>test</scope>
+		</dependency>
+	</dependencies>
+</project>
```
cppwfs: that evil last line.  Will fix on merge.

spring-cloud-task-stream/src/test/resources/org/springframework/cloud/task/listener/sink-channel.properties
```diff
@@ -0,0 +1 @@
+spring.cloud.stream.bindings.input.destination=test
```
cppwfs: We can remove this file it is no longer needed here.  Will fix on merge.

cppwfs:
LGTM

cppwfs:
Squashed Merged a little cleanup.  Thanks for doing this Ilaya!

PR #136, opened 2016-05-06
Fix autoconfiguration class entries

### LINKED ISSUE #135: Task event channel binding doesn't happen
I have a task with the `spring-cloud-task-stream` dependency. When the task is launched, I only see the batch event channel registration and `task-event` channel registration doesn't happen.

It looks like the `TaskEventAutoConfiguration` is never invoked.

ilayaperumalg:
- This will make sure to invoke all the auto configuration classes

Resolves #135


cppwfs:
LGTM

Issue #139, opened 2016-05-10
Task not recording exit code nor message on uncaught exception


cppwfs:
Tasks should report the exit code of 1 (if not exit genererator is present) and the exception info if the task ends because of an uncaught exception.

PR #140, opened 2016-05-10
SCT-139 Exception info and error codes will be recorded for the task

cppwfs:
resolves spring-cloud/spring-cloud-task#139


mminella:
I may be missing something but I don't see how the test updates in this PR actually validate the updated condition.  They look like they are "fixed" in that they don't break anymore, but I don't see how they validate that the listener is called (the condition that caused this fix to be needed in the first place).

cppwfs:
Agreed.  I will add new tests.

mminella:
LGTM.  Merged as df14546a7fce5cb58761965e0c4b1347c27ff60a

Issue #146, opened 2016-05-23
TaskLaunchRequest should use URI vs maven structure


cppwfs:
Upgrade TaskLaunchRequest to use URI to reference artifacts

PR #147, opened 2016-05-24
Update TaskLaunchRequest to use URI

cppwfs:
resolves spring-cloud/spring-cloud-task#146

### DIFF
--- a/spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherConfiguration.java
+++ b/spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherConfiguration.java
@@ -16,13 +16,21 @@
 
 package org.springframework.cloud.task.launcher;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.cloud.deployer.resource.maven.MavenProperties;
+import org.springframework.cloud.deployer.resource.maven.MavenResourceLoader;
+import org.springframework.cloud.deployer.resource.support.DelegatingResourceLoader;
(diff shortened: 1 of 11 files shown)


spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherConfiguration.java
```diff
@@ -44,4 +52,25 @@ public TaskLauncher taskLauncher() {
...
+	}
+
+	@Bean
+	public MavenProperties mavenProperties() {
+		return new MavenConfigurationProperties();
+	}
+
+	@ConfigurationProperties(prefix = "maven")
+	static class MavenConfigurationProperties extends MavenProperties {
+	}
```
mminella: Isn't all of this really dependent on the launcher?  AKA if we swapped the K8s launcher for the local task launcher, we wouldn't need this maven stuff, we'd need docker stuff, right?

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
```diff
@@ -55,15 +60,9 @@ public void taskLauncherSink(TaskLaunchRequest request) {
...
-				.artifactId(taskLaunchRequest.getArtifact())
-				.groupId(taskLaunchRequest.getTaskGroupId())
-				.version(taskLaunchRequest.getTaskVersion())
-				.extension(taskLaunchRequest.getTaskExtension())
-				.classifier(taskLaunchRequest.getTaskClassifier())
-				.build();
-		AppDefinition definition = new AppDefinition(taskLaunchRequest.getArtifact(), taskLaunchRequest.getProperties());
-		AppDeploymentRequest request = new AppDeploymentRequest(definition, resource);
+		Resource resource = delegatingResourceLoader.getResource(taskLaunchRequest.getUri());
+		AppDefinition definition = new AppDefinition("Task-" + taskLaunchRequest.hashCode(), taskLaunchRequest.getProperties());
```
mminella: I think we can do better for a name than just `Task-<SOME_MEANINGLESS_STRING>`...

PR #148, opened 2016-06-08
Refactored the use of the name parameters to be arguments

mminella:
This PR refactors the use of the term parameters to arguments to be
consistent with how the component is utilized within Spring Cloud Data
Flow.

### DIFF
--- a/spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDaoTests.java
+++ b/spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDaoTests.java
@@ -57,9 +57,9 @@ public class JdbcTaskExecutionDaoTests {
 	@Test
 	@DirtiesContext
 	public void saveTaskExecution() {
-		TaskExecution expectedTaskExecution = TestVerifierUtils.createSampleTaskExecutionNoParam();
+		TaskExecution expectedTaskExecution = TestVerifierUtils.createSampleTaskExecutionNoArg();
 		expectedTaskExecution = dao.createTaskExecution(expectedTaskExecution.getTaskName(), expectedTaskExecution.getStartTime(),
-				expectedTaskExecution.getParameters());
+				expectedTaskExecution.getArguments());
 
 		TestVerifierUtils.verifyTaskExecution(expectedTaskExecution,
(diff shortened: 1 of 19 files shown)


spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDaoTests.java
```diff
@@ -70,7 +70,7 @@ public void saveTaskExecution() {
 	public void completeTaskExecution() {
 		TaskExecution expectedTaskExecution = TestVerifierUtils.endSampleTaskExecutionNoParam();
```
cppwfs: Rename test from NoParam to NoArg.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDaoTests.java
```diff
@@ -48,7 +48,7 @@ public void completeTaskExecution(){
 		MapTaskExecutionDao dao = new MapTaskExecutionDao();
 		TaskExecution expectedTaskExecution = TestVerifierUtils.createSampleTaskExecutionNoParam();
```
cppwfs: Rename test from NoParam to NoArg.

cppwfs:
In the TestVerifierUtils  class replace the method signatures with the term NoParam with noArg.

cppwfs:
In the TestVerifierUtils class replace PARAM_SIZE with ARG_SIZE

cppwfs:
In the TestVerifierUtils class replace assert messages that contain the word `parameters` with the word `arguments`

cppwfs:
Looking at the createTaskExecution method in the TaskExecutionDao interface.   It accepts a `parameters`  parameter.  Since it only populates the arguments attribute of the task execution, we probably need to change it to arguments.  And cascade this down the the Impl's.  Thoughts?

cppwfs:
Looking at the createTaskExecution method in the TaskRepository interface. It accepts a parameters parameter. Since it only populates the arguments attribute of the task execution, we probably need to change it to arguments. And cascade this down the the Impl's. Thoughts?

mminella:
Updated per your comments.

cppwfs:
LGTM Merged.

PR #152, opened 2016-06-14
Updated versions in documentation & dependency management within samples

mminella:
Versions where updated in documentation that is included in the
generation of our reference documentation.

Also, to clean up version management within the samples, the
spring-cloud-task-dependencies was introduced into each pom so that
versions are managed at that level.

### DIFF
--- a/spring-cloud-task-samples/task-events/pom.xml
+++ b/spring-cloud-task-samples/task-events/pom.xml
@@ -22,6 +22,18 @@
 		<java.version>1.7</java.version>
 	</properties>
 
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-task-dependencies</artifactId>
+				<version>1.0.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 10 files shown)


spring-cloud-task-samples/task-events/pom.xml
```diff
@@ -31,19 +43,17 @@
...
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-task-stream</artifactId>
-			<version>1.0.0.BUILD-SNAPSHOT</version>
 		</dependency>
 
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-starter-stream-rabbit</artifactId>
-			<version>1.0.2.RELEASE</version>
+			<version>1.0.3.BUILD-SNAPSHOT</version>
```
cppwfs: Why are we using the snapshot vs the release?

cppwfs: And this can be added to dependencies.

cppwfs:
Updated one version number, squashed merged.

PR #155, opened 2016-06-24
Document how Job Failures relate to Tasks

### LINKED ISSUE #123: Document how job failures relate to Task executions
When running Batch Jobs using SCT the behavior is if *Job fails* the corresponding *Task succeeds*.

Users may have a different expectation so that if *Job fails* then *Task fails*.

In that case user would have to write their own `CommandLineRunner` that bubbles the error up.

The boot one that is used by default doesn’t. 

Therefore, we need to add relevant reference documentation.

cppwfs:
Resolves #123

### DIFF
--- a/spring-cloud-task-docs/src/main/asciidoc/batch.adoc
+++ b/spring-cloud-task-docs/src/main/asciidoc/batch.adoc
@@ -107,3 +107,18 @@ public DeployerStepExecutionHandler stepExecutionHandler(JobExplorer jobExplorer
 Spring Cloud Task provides the ability for batch jobs to emit informational messages.  This
 is covered in detail in the section
 <<stream.adoc#stream-integration-batch-events, Spring Batch Events>>.
+
+[[batch-failures-and-tasks]]
+== Batch Job Exit Codes
+
+As discussed before Spring Cloud Task applications support the ability to record the exit code of a task execution.
+However in cases where a user is running a Spring Batch Job within a task, regardless of how the Batch Job
+Execution completes the result of the task will always be zero when using a CommandLineRunner for example
+(unless there is an uncaught exception).  Keep in mind
+that a task is a boot application and the exit code returned from the task is the same as a boot application.  So
+to have your task return the exit code based on the result of the batch job execution, you can take advantage of the
+http://docs.spring.io/spring-batch/reference/html/configureJob.html#commandLineJobRunner[CommandLineJobRunner] or
+write your own code that will check the result of the batch job execution and take advantage of Spring Boot's
+http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-application-exit[ExitCodeGenerator]
+interface.
+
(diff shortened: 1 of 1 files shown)


spring-cloud-task-docs/src/main/asciidoc/batch.adoc
```diff
@@ -107,3 +107,18 @@ public DeployerStepExecutionHandler stepExecutionHandler(JobExplorer jobExplorer
...
+As discussed before Spring Cloud Task applications support the ability to record the exit code of a task execution.
+However in cases where a user is running a Spring Batch Job within a task, regardless of how the Batch Job
+Execution completes the result of the task will always be zero when using a CommandLineRunner for example
+(unless there is an uncaught exception).  Keep in mind
+that a task is a boot application and the exit code returned from the task is the same as a boot application.  So
+to have your task return the exit code based on the result of the batch job execution, you can take advantage of the
+http://docs.spring.io/spring-batch/reference/html/configureJob.html#commandLineJobRunner[CommandLineJobRunner] or
+write your own code that will check the result of the batch job execution and take advantage of Spring Boot's
+http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-application-exit[ExitCodeGenerator]
+interface.
```
mminella: This isn't correct.  A couple issues:

* You can't use the `CommandLineJobRunner` because it's not a `CommandLineRunner` (it pre-dates boot) and doesn't integrate with boot.
* The framework isn't going to let an uncaught exception out from the batch job so using the `ExitCodeGenerator` won't be of much help.

Really, if you want the task to fail if the job fails, right now you need to write your own `CommandLineRunner`.

cppwfs: * good catch. I'll update to remove  the CommandLineJobRunner.  
* For this I was thinking outside of the Batch app (i.e. just the boot application).  But in this case I agree let's focus on the batch job.

spring-cloud-task-docs/src/main/asciidoc/batch.adoc
```diff
@@ -107,3 +107,14 @@ public DeployerStepExecutionHandler stepExecutionHandler(JobExplorer jobExplorer
 Spring Cloud Task provides the ability for batch jobs to emit informational messages.  This
 is covered in detail in the section
 <<stream.adoc#stream-integration-batch-events, Spring Batch Events>>.
+
+[[batch-failures-and-tasks]]
+== Batch Job Exit Codes
+
+As discussed before Spring Cloud Task applications support the ability to record the exit code of a task execution.
+However in cases where a user is running a Spring Batch Job within a task, regardless of how the Batch Job
+Execution completes the result of the task will always be zero when using a CommandLineRunner for example.  Keep in mind
```
mminella: It won't be zero when using "a CommandLineRunner"...it will be zero when using the default Batch and Boot functionality.  I can change this on merge.

cppwfs: Agreed.

mminella:
Merged as 5393962685ea9d6005300b9f08570f81ab9448b4

PR #156, opened 2016-06-27
Replaced the mysql dependency with mariadb

cppwfs:
To remain standard with the other applications.
Also updated documentation.


mminella:
Merged as eddd38797b2859a30eda624ce04ca43b0be23865 and 3ec0769e0252d13ad9f9536d5f85aa27fd2bc15c

PR #161, opened 2016-06-29
Updates the SCT to use the latest deployer 1.0.1.RELEASE

cppwfs:
And updates the MavenResource used by TaskLauncherSink

### DIFF
--- a/spring-cloud-task-docs/src/main/asciidoc/batch.adoc
+++ b/spring-cloud-task-docs/src/main/asciidoc/batch.adoc
@@ -68,7 +68,7 @@ public PartitionHandler partitionHandler(TaskLauncher taskLauncher,
 		MavenResource.parse(String.format("%s:%s:%s",
 				"io.spring.cloud",
 				"partitioned-batch-job",
-				"1.0.0.RELEASE"));
+				"1.0.1.RELEASE"));
 
 	DeployerPartitionHandler partitionHandler =
 		new DeployerPartitionHandler(taskLauncher, jobExplorer, resource, "workerStep");
--- a/spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
+++ b/spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
@@ -16,9 +16,14 @@
 
 package org.springframework.cloud.task.launcher;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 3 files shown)


spring-cloud-task-docs/src/main/asciidoc/batch.adoc
```diff
@@ -68,7 +68,7 @@ public PartitionHandler partitionHandler(TaskLauncher taskLauncher,
 		MavenResource.parse(String.format("%s:%s:%s",
```
markfisher: I think you should show the passing of `MavenProperties` here, with `remoteRepositories` configured (which would typically point to the repo where the custom batch apps are published). Otherwise, this would only be relying upon the local repo.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
```diff
@@ -60,7 +65,10 @@ public void taskLauncherSink(TaskLaunchRequest request) {
 	private void launchTask(TaskLaunchRequest taskLaunchRequest) {
 		Assert.notNull(taskLauncher, "TaskLauncher has not been initialized");
 		logger.info("Launching Task for the following resource " + taskLaunchRequest);
-		Resource resource = delegatingResourceLoader.getResource(taskLaunchRequest.getUri());
+		MavenProperties mavenProperties = new MavenProperties();
+		Map<String, MavenProperties.RemoteRepository> remoteRepositoryMap = new HashMap<>();
```
markfisher: did you intend to add a repo entry here (something would need to be configurable as a property of the launcher app itself)?... otherwise this cannot retrieve an app from anywhere except "local" and this app could very likely be running in an environment that has nothing published to its local repo

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
```diff
@@ -60,7 +65,10 @@ public void taskLauncherSink(TaskLaunchRequest request) {
 	private void launchTask(TaskLaunchRequest taskLaunchRequest) {
 		Assert.notNull(taskLauncher, "TaskLauncher has not been initialized");
 		logger.info("Launching Task for the following resource " + taskLaunchRequest);
-		Resource resource = delegatingResourceLoader.getResource(taskLaunchRequest.getUri());
```
markfisher: actually I think you would still want to use the resource loader here... you just need to make sure that the `MavenResourceLoader` that is configured within the `DelegatingResourceLoader` has `remoteRepositories` provided on its `MavenProperties`

cppwfs:
Closing en-leu of using https://github.com/spring-cloud/spring-cloud-task/pull/162 instead.

PR #162, opened 2016-06-30
Update SCT to use 1.0.1.RELEASE of deployers

cppwfs:
Update the TaskLauncherConfiguration and sample partitioner to use a configurable remote repository.
Updated readme's to 1.0.1 for tasks
Updated the docs for configuring remote repostiory.

### DIFF
--- a/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchAutoConfiguration.java
+++ b/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchAutoConfiguration.java
@@ -26,6 +26,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.deployer.resource.maven.MavenProperties;
@@ -33,6 +34,7 @@ import org.springframework.cloud.deployer.resource.maven.MavenResourceLoader;
 import org.springframework.cloud.deployer.resource.support.DelegatingResourceLoader;
 import org.springframework.cloud.task.batch.listener.TaskBatchExecutionListener;
 import org.springframework.cloud.task.repository.TaskExplorer;
+import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.io.ResourceLoader;
@@ -47,6 +49,9 @@ import org.springframework.util.CollectionUtils;
 @ConditionalOnBean({Job.class})
 @ConditionalOnProperty(name = "spring.cloud.task.batch.listener.enable", havingValue = "true", matchIfMissing = true)
 public class TaskBatchAutoConfiguration {
+
+	@Autowired
+	ApplicationContext context;
 
 	@Bean
 	@ConditionalOnMissingBean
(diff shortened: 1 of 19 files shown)


spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherConfiguration.java
```diff
@@ -42,6 +44,9 @@
 @ConditionalOnClass({TaskLauncher.class})
 public class TaskLauncherConfiguration {
 
+	@Value("${maven.remoteRepositories.springRepo.url:https://repo.spring.io/libs-snapshot}")
+	private String repository;
```
markfisher: Why add this property when you could have just relied on the `MavenProperties` below which were exposed as boot properties? For one thing, if someone needs to set auth or proxy settings, they would need those as well, and now that you've removed `@ConfigurationProperties` there, those other properties are no longer available.

spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
```diff
@@ -91,7 +97,12 @@ public TaskLauncher taskLauncher() {
 
 	@Bean
 	public PartitionHandler partitionHandler(TaskLauncher taskLauncher, JobExplorer jobExplorer) throws Exception {
-		MavenResource resource = MavenResource.parse("io.spring.cloud:partitioned-batch-job:1.0.0.BUILD-SNAPSHOT");
+
```
markfisher: if you use the `delegatingResourceLoader` here, you don't need to control the maven properties here (since they'd be applied to the `MavenResourceLoader` within the `delegatingResourceLoader` *and* you would get the benefit of being able to load from more than Maven, e.g. `file:` or `http:` URLs.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherConfiguration.java
```diff
@@ -16,9 +16,11 @@
 
 package org.springframework.cloud.task.launcher;
 
+import java.util.Collections;
```
markfisher: don't need this import

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherConfiguration.java
```diff
@@ -16,9 +16,11 @@
...
 import java.util.HashMap;
 import java.util.Map;
 
+import org.springframework.beans.factory.annotation.Value;
```
markfisher: don't need this import

spring-cloud-task-docs/src/main/asciidoc/stream.adoc
```diff
@@ -42,6 +42,10 @@ repository execute a maven build from the `spring-cloud-task-samples` directory
 property `skipInstall` set to false.  For example:
 `mvn clean install`.
 
+NOTE:  The maven.remoteRepositories.springRepo.url property will need to be set to
+the location of the remote repository from which the über-jar is located.  If not
+set, the default will use the local repo.
```
markfisher: should probably be more precise... by default there will be no remote repository, so it will rely upon the local repository only

spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
```diff
@@ -91,7 +97,12 @@ public TaskLauncher taskLauncher() {
...
+		MavenProperties mavenProperties = new MavenProperties();
+		mavenProperties.setRemoteRepositories(new HashMap<>(Collections.singletonMap("springRepo",
```
mminella: Why the `new HashMap` here?

spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
cppwfs: Resolved this in latest update.

spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
```diff
@@ -47,13 +50,19 @@
...
 
 /**
  * @author Michael Minella
+ * @author Glenn Renfro
  */
 @Configuration
 public class JobConfiguration {
 
+	@Value("${maven.remoteRepositories.springRepo.url:https://repo.spring.io/libs-snapshot}")
+	private String repository;
```
markfisher: this is no longer used, right?

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchAutoConfiguration.java
```diff
@@ -66,4 +74,31 @@ else if(CollectionUtils.isEmpty(dataSources)) {
...
+	@Configuration
+	@ConditionalOnClass(name = "org.springframework.cloud.deployer.resource.maven.MavenResourceLoader")
+	public static class MavenResourceConfiguration {
+		@Bean
+		public MavenResourceLoader mavenResourceLoader(MavenProperties properties) {
+			return new MavenResourceLoader(properties);
+		}
+
+		@Bean
+		@ConditionalOnMissingBean(DelegatingResourceLoader.class)
```
markfisher: if this config is not included (the condition above is not satisfied), then wouldn't there need to be a `DelegatingResourceLoader` that does not include the `MavenResourceLoader` - since I believe that is expected as an autowired property?

cppwfs:
Closing this PR.  A refactoring effort will replace this code.

PR #163, opened 2016-07-01
Refactored dependency management

mminella:
Created a Spring Boot starter that can be used to configure Spring Cloud
Task and it's related functionality.

### DIFF
--- a/spring-cloud-task-batch/pom.xml
+++ b/spring-cloud-task-batch/pom.xml
@@ -27,7 +27,13 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-deployer-local</artifactId>
+			<optional>true</optional>
 			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-deployer-resource-support</artifactId>
+			<optional>true</optional>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.batch</groupId>
@@ -43,6 +49,7 @@
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-test</artifactId>
+			<scope>test</scope>
 		</dependency>
 		<dependency>
 			<groupId>junit</groupId>
--- a/spring-cloud-task-batch/pom.xml
+++ b/spring-cloud-task-batch/pom.xml
@@ -27,7 +27,6 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-deployer-local</artifactId>
-			<optional>true</optional>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
--- a/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
+++ b/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
@@ -543,6 +543,58 @@ public class DeployerPartitionHandlerTests {
 		validateStepExecutionResults(results);
 	}
 
+	@Test
+	public void testDeployerProperties() throws Exception {
+
+		StepExecution masterStepExecution = createMasterStepExecution();
+		JobExecution jobExecution = masterStepExecution.getJobExecution();
+
+		StepExecution workerStepExecutionStart = getStepExecutionStart(jobExecution, 4L);
+		StepExecution workerStepExecutionFinish = getStepExecutionFinish(workerStepExecutionStart, BatchStatus.COMPLETED);
+
+		DeployerPartitionHandler handler = new DeployerPartitionHandler(this.taskLauncher, this.jobExplorer, this.resource, "step1");
+		handler.setEnvironment(this.environment);
+
+		Map<String, String> deploymentProperties = new HashMap<>(2);
+		deploymentProperties.put("foo", "bar");
+		deploymentProperties.put("baz", "qux");
+
+		handler.setDeploymentProperties(deploymentProperties);
+
+		TaskExecution taskExecution = new TaskExecution();
+		taskExecution.setTaskName("partitionedJobTask");
+
+		Set<StepExecution> stepExecutions = new HashSet<>();
+		stepExecutions.add(workerStepExecutionStart);
+		when(this.splitter.split(masterStepExecution, 1)).thenReturn(stepExecutions);
+
+		when(this.jobExplorer.getStepExecution(1L, 4L)).thenReturn(workerStepExecutionFinish);
+
+		handler.beforeTask(taskExecution);
+		Collection<StepExecution> results = handler.handle(this.splitter, masterStepExecution);
+
+		verify(this.taskLauncher).launch(this.appDeploymentRequestArgumentCaptor.capture());
+
+		AppDeploymentRequest request = this.appDeploymentRequestArgumentCaptor.getValue();
+
+		assertEquals(this.resource, request.getResource());
+		assertEquals(2, request.getDeploymentProperties().size());
+		assertEquals("bar", request.getDeploymentProperties().get("foo"));
+		assertEquals("qux", request.getDeploymentProperties().get("baz"));
+
+		AppDefinition appDefinition = request.getDefinition();
+
+		assertEquals("partitionedJobTask:partitionedJob:step1:partition1", appDefinition.getName());
+		assertTrue(request.getCommandlineArguments().contains(formatArgs(DeployerPartitionHandler.SPRING_CLOUD_TASK_JOB_EXECUTION_ID, "1")));
+		assertTrue(request.getCommandlineArguments().contains(formatArgs(DeployerPartitionHandler.SPRING_CLOUD_TASK_STEP_EXECUTION_ID, "4")));
+		assertTrue(request.getCommandlineArguments().contains(formatArgs(DeployerPartitionHandler.SPRING_CLOUD_TASK_STEP_NAME, "step1")));
+
+		assertEquals(1, results.size());
+		StepExecution resultStepExecution = results.iterator().next();
+		assertEquals(BatchStatus.COMPLETED, resultStepExecution.getStatus());
+		assertEquals("step1:partition1", resultStepExecution.getStepName());
+	}
+
 	private StepExecution getStepExecutionFinish(StepExecution stepExecutionStart, BatchStatus status) {
 		StepExecution workerStepExecutionFinish = new StepExecution(stepExecutionStart.getStepName(), stepExecutionStart.getJobExecution());
 		workerStepExecutionFinish.setId(stepExecutionStart.getId());
--- a/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
+++ b/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
@@ -147,9 +147,9 @@ public class DeployerPartitionHandlerTests {
 		AppDefinition appDefinition = request.getDefinition();
 
 		assertEquals("partitionedJobTask:partitionedJob:step1:partition1", appDefinition.getName());
-		assertEquals("1", appDefinition.getProperties().get(DeployerPartitionHandler.SPRING_CLOUD_TASK_JOB_EXECUTION_ID));
-		assertEquals("4", appDefinition.getProperties().get(DeployerPartitionHandler.SPRING_CLOUD_TASK_STEP_EXECUTION_ID));
-		assertEquals("step1", appDefinition.getProperties().get(DeployerPartitionHandler.SPRING_CLOUD_TASK_STEP_NAME));
+		assertTrue(request.getCommandlineArguments().contains(formatArgs(DeployerPartitionHandler.SPRING_CLOUD_TASK_JOB_EXECUTION_ID, "1")));
+		assertTrue(request.getCommandlineArguments().contains(formatArgs(DeployerPartitionHandler.SPRING_CLOUD_TASK_STEP_EXECUTION_ID, "4")));
+		assertTrue(request.getCommandlineArguments().contains(formatArgs(DeployerPartitionHandler.SPRING_CLOUD_TASK_STEP_NAME, "step1")));
 
 		assertEquals(1, results.size());
 		StepExecution resultStepExecution = results.iterator().next();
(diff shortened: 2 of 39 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.io.ResourceLoader;
+
+/**
+ * Autoconfiguration of a file or Maven based {@link ResourceLoader}.
+ *
+ * @author Michael Minella
+ * @since 1.0.2
```
cppwfs: 1.0.1

spring-cloud-task-docs/src/main/asciidoc/batch.adoc
```diff
@@ -85,11 +90,12 @@ public PartitionHandler partitionHandler(TaskLauncher taskLauncher,
...
-used as the job repository and task repository.  Once the underlying infrastructure has
-bootstrapped the Spring Boot jar and Spring Boot has launched the
-`DeployerStepExecutionHandler`, the step handler will execute the Step requested.  An
-example of configuring the `DefaultStepExecutionHandler`:
+The repository enumerated in the example above should be the location of the remote repository
+from which the über-jar is located. Both the master and slave are expected to have
+visibility into the same data store being used as the job repository and task repository.
+Once the underlying infrastructure has bootstrapped the Spring Boot jar and Spring Boot
+has launched the `DeployerStepExecutionHandler`, the step handler will execute the Step
+requested.  An example of configuring the `DefaultStepExecutionHandler`:
```
cppwfs: An example of configuring the `DefaultStepExecutionHandler` is shown below:

spring-cloud-task-batch/pom.xml
```diff
@@ -27,9 +27,15 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-deployer-local</artifactId>
+			<optional>true</optional>
```
cppwfs: argue with me.  But we may not need the optional here if scope is test.

spring-cloud-task-integration-tests/pom.xml
```diff
@@ -36,5 +36,14 @@
...
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.batch</groupId>
+			<artifactId>spring-batch-core</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-deployer-resource-support</artifactId>
+		</dependency>
```
cppwfs: We may want to exclude integrations tests  from the main set of modules.  And create a maven profile for running integration tests.

spring-cloud-task-batch/pom.xml
mminella: No need to argue when you're right.

spring-cloud-task-integration-tests/pom.xml
mminella: Agreed.  We shouldn't be publishing them either.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
```diff
@@ -0,0 +1,86 @@
...
+ * Autoconfiguration of a file or Maven based {@link ResourceLoader}.
+ *
+ * @author Michael Minella
+ * @since 1.0.1
+ */
+@Configuration
+public class ResourceLoadingAutoConfiguration {
+
+	@Configuration
+	@ConditionalOnClass(MavenResourceLoader.class)
```
markfisher: should the value be a String instead of the Class (since it might not be present)?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
```diff
@@ -0,0 +1,86 @@
...
+		@Autowired
+		private ApplicationContext context;
+
+		@Bean
+		@ConditionalOnMissingBean
+		public DelegatingResourceLoader delegatingResourceLoader() {
+			Map<String, ResourceLoader> loaders = new HashMap<>(1);
+			loaders.put("file", this.context);
+
+			return new DelegatingResourceLoader(loaders);
```
markfisher: should we go ahead and change the `DelegatingResourceLoader

spring-cloud-task-docs/src/main/asciidoc/batch.adoc
```diff
@@ -64,11 +64,16 @@ look like the following:
 @Bean
 public PartitionHandler partitionHandler(TaskLauncher taskLauncher,
 		JobExplorer jobExplorer) throws Exception {
-	MavenResource resource =
+
+	MavenProperties mavenProperties = new MavenProperties();
```
markfisher: would it be too much for this example to show the use of `DelegatingResourceLoader` instead of this way of handling the `MavenProperties`?

spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
```diff
@@ -69,6 +71,9 @@
 	@Autowired
 	private ConfigurableApplicationContext context;
 
+	@Autowired
```
markfisher: I feel like this should be a param for the `partitionHandler` bean config instead.

Or  the `@Autowired` could contain `required=false` so that if the conditional config does not find the `MavenResourceLoader`, no other `DelegatingResourceLoader` bean needs to be defined. Instead, a `DefaultResourceLoader` could be used for the `partitionHandler` if the autowired field is null.

spring-cloud-task-starter/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+	<name>Spring Cloud Task Starter</name>
+	<description>Spring Boot starter for Spring Cloud Task</description>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-task-core</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
```
markfisher: Adding `batch` and `stream` makes this a pretty heavy starter. Do we foresee anyone wanting a boot starter for very simple task apps?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
```diff
@@ -0,0 +1,86 @@
...
+		@ConfigurationProperties(prefix = "maven")
+		public static class MavenConfigurationProperties extends MavenProperties {}
+	}
+
+	@Configuration
+	@ConditionalOnMissingClass("org.springframework.cloud.deployer.resource.maven.MavenResourceLoader")
+	public static class LocalResourceLoadingAutoConfiguration {
+
+		@Autowired
+		private ApplicationContext context;
```
markfisher: why is this autowired?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
mminella: Per the javadoc, it should be safe either way (unless I'm reading it wrong): http://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/condition/ConditionalOnClass.html

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
mminella: That would be very nice...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
markfisher: okay, well we should probably be consistent within this config class in any case

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
markfisher: I think for now, my other suggestion (required=false for autowiring) would be ok

spring-cloud-task-docs/src/main/asciidoc/batch.adoc
mminella: I think that's fine...

spring-cloud-task-starter/pom.xml
mminella: This is based on a conversation with @philwebb.  Both those dependencies have everything else optional so they actually should be very light additions.  This approach also allows you to add this starter with the appropriate starter (say batch or stream starters) and the autoconfig would kick in.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
mminella: Meaning?  `@ConditionalOnClass` allows this option.  `@ConditionaOnMissingClass` requires a `String` (no `Class` option).

markfisher: gotcha - nevermind

spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
mminella: That only works with the above change (not requiring a resource loader), right?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
mminella: Vs injecting in the method below?  Doesn't really matter either way as far as I'm concerned...do we have an official preference?

philwebb: FWIW we now use constructor injection in Spring Boot `@Configuration` classes.

spring-cloud-task-dependencies/pom.xml
```diff
@@ -17,10 +16,12 @@
 
 	<properties>
 		<spring-cloud-stream.version>1.0.2.RELEASE</spring-cloud-stream.version>
-		<spring-cloud-deployer-spi.version>1.0.0.RELEASE</spring-cloud-deployer-spi.version>
-		<spring-cloud-deployer-local.version>1.0.0.RELEASE</spring-cloud-deployer-local.version>
+		<spring-cloud-deployer-spi.version>1.0.1.RELEASE</spring-cloud-deployer-spi.version>
```
cppwfs: Update the version to 1.0.2.BUILD-SNAPSHOT.

cppwfs:
there is a breaking change when we go from 1.0.1 to 1.0.2 of the dependencies.  The datasource properties are not being cascaded to the partitions.

mminella:
How are you configuring the datasource?

cppwfs:
Via environment variable.  Worked with 1.0.1.  
If I add `localDeployerProperties.setEnvVarsToInherit(new String[]{"spring.datasource.url", "spring.datasource.username","spring.datasource.password", "spring.datasource.driverClassName"});`
in TaskLauncher bean definition in the partition-job-sample they do get propagated.  Then the other problem is that the first 2 partitions do fire and complete but the 2 partitions don't get started.

spring-cloud-task-docs/src/main/asciidoc/batch.adoc
markfisher: actually could just use `MavenResourceLoader` directly... but I guess what's shown now is fine - it's a bit lower level though

spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
markfisher: no - I'm just saying that the field could be null if no `DelegatingResourceLoader` is defined in the context... in that case, a `DefaultResourceLoader` could be used directly

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
```diff
@@ -344,23 +344,27 @@ public void testPassingEnvironmentProperties() throws Exception {
...
+		assertEquals("bar", request.getDefinition().getProperties().get("foo"));
+		assertEquals("qux", request.getDefinition().getProperties().get("baz"));
 
 		AppDefinition appDefinition = request.getDefinition();
 
 		assertEquals("partitionedJobTask:partitionedJob:step1:partition1", appDefinition.getName());
-		assertEquals("1", appDefinition.getProperties().get(DeployerPartitionHandler.SPRING_CLOUD_TASK_JOB_EXECUTION_ID));
-		assertEquals("4", appDefinition.getProperties().get(DeployerPartitionHandler.SPRING_CLOUD_TASK_STEP_EXECUTION_ID));
-		assertEquals("step1", appDefinition.getProperties().get(DeployerPartitionHandler.SPRING_CLOUD_TASK_STEP_NAME));
+		assertTrue(request.getCommandlineArguments().contains(formatArgs(DeployerPartitionHandler.SPRING_CLOUD_TASK_JOB_EXECUTION_ID, "1")));
```
markfisher: technically the *expected* value should always be the first parameter and the value under test second

spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
mminella: Tasks provides one by default.  See the `ResourceLoadingAutoConfiguration`.  We should always have one.

markfisher: I know. What I am trying to suggest is that the `DelegatingResourceLoader` is only necessary when the Maven condition is satisfied. Otherwise a `DefaultResourceLoader` is sufficient. As is, this bean is really useless: https://github.com/spring-cloud/spring-cloud-task/pull/163/files#diff-fb5d1b3d21fb4b4c279d5591934717afR73

...and that is the only bean in that configuration class, so that one could be avoided altogether, leaving only the one that is conditional upon the Maven support.

Hope that makes sense.

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
mminella: That line is an `assertTrue`.  There is no expected value...

markfisher: duh sorry - and all of the `assertEquals` look correct... I was blurring 2 sides of the diff in my mind I guess... carry on ;)

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLaunchRequest.java
```diff
@@ -37,21 +37,24 @@
...
-	private Map<String, String> properties;
+	private Map<String, String> environmentProperties;
+	private Map<String, String> deploymentProperties;
 
 	/**
 	 * Constructor for the TaskLaunchRequest;
 	 * @param uri the URI to the task artifact to be launched.
 	 * @param commandlineArguments list of commandlineArguments to be used by the task
-	 * @param properties is the environment variables for this task.
+	 * @param environmentProperties is the environment variables for this task.
```
markfisher: should add `deploymentProperties` in the javadoc

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLaunchRequest.java
```diff
@@ -103,15 +116,15 @@ public boolean equals(Object o) {
 		if (!(commandlineArguments != null ? commandlineArguments.equals(that.commandlineArguments) : that.commandlineArguments == null)){
 			return false;
 		}
-		return properties != null ? properties.equals(that.properties) : that.properties == null;
+		return environmentProperties != null ? environmentProperties.equals(that.environmentProperties) : that.environmentProperties == null;
```
markfisher: `deploymentProperties` should be included in `equals` and `hashCode`

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLaunchRequest.java
cppwfs: Will fix on merge

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLaunchRequest.java
cppwfs: will fix on merge

cppwfs:
LGTM 
Fixed last 2 code review requests on merge
Squash, Small fixes, Merge.

PR #168, opened 2016-07-06
Removed unused default `DelegatingResourceLoader`

mminella:


cppwfs:
LGTM

cppwfs:
Spoke too soon.  When running the sample it is not finding the maven://io.spring.cloud:partitioned-batch-job:1.0.1.BUILD-SNAPSHOT in the maven repo.

mminella:
What's the command you're running?  It needs to be installed in your maven repo in order for that to work...

spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
```diff
@@ -72,7 +70,7 @@
 	private ConfigurableApplicationContext context;
 
 	@Autowired
-	private DelegatingResourceLoader delegatingResourceLoader;
+	private ResourceLoader resourceLoader;
```
markfisher: I'm surprised this would work, considering there should be more than 1 autowire candidate.

cppwfs:
Closed en-leu of https://github.com/spring-cloud/spring-cloud-task/pull/169

PR #169, opened 2016-07-07
Updated to use DelegatedResourceLoader that can use an empty loaders map

cppwfs:
Replaces PR https://github.com/spring-cloud/spring-cloud-task/pull/168

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
@@ -72,11 +72,8 @@ public class ResourceLoadingAutoConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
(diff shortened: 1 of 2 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfiguration.java
```diff
@@ -72,11 +72,8 @@ public MavenProperties mavenProperties() {
 
 		@Bean
 		@ConditionalOnMissingBean
-		public DelegatingResourceLoader delegatingResourceLoader(ApplicationContext context) {
-			Map<String, ResourceLoader> loaders = new HashMap<>(1);
-			loaders.put("file", context);
-
-			return new DelegatingResourceLoader(loaders);
+		public DelegatingResourceLoader delegatingResourceLoader() {
+			return new DelegatingResourceLoader(new HashMap<String, ResourceLoader>());
```
markfisher: should be possible to pass `null` instead of creating the empty map here

mminella:
LGTM.  Merged as de3d09664e7076e11bdf6b644155809fc44416fc

Issue #181, opened 2016-07-22
Add environment variable customization to `DeployerPartitionHandler`


mminella:
In order to support more robust configuration of the environment variables provided to the workers of a partitioned job, we need to add an abstraction around how they are configured.

PR #182, opened 2016-07-22
Added abstraction for defining environment variables

mminella:
When launching workers as separate tasks, it can be useful to be able to
use additional logic to define environment variables.  This commit
provides an abstraction to allow for the customization of environment
variables on a per worker basis as well as two useful implementations:

* A no-op implementation (returns an empty Map).
* An implementation that moves the current environment variable handling
* out of the `DeployerPartitionHandler`.

Resolves spring-cloud/spring-cloud-task#181

### DIFF
--- a/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/NoOpEnvironmentVariablesProvider.java
+++ b/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/NoOpEnvironmentVariablesProvider.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.task.batch.partition;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.batch.item.ExecutionContext;
+
+/**
+ * A simple no-op implementation of the {@link EnvironmentVariablesProvider}.  It returns
+ * an empty {@link Map}.
+ *
+ * @author Michael Minella
+ *
+ * @since 1.0.2
+ */
+public class NoOpEnvironmentVariablesProvider implements EnvironmentVariablesProvider {
+
+	/**
+	 *
+	 * @param executionContext the {@link ExecutionContext} associated with the worker's
+	 * 			step
+	 * @return an empty {@link Map}
+	 */
+	@Override
+	public Map<String, String> getEnvironmentVariables(ExecutionContext executionContext) {
+		return new HashMap<>(0);
+	}
+}
--- a/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/SimpleEnvironmentVariablesProvider.java
+++ b/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/SimpleEnvironmentVariablesProvider.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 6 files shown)


spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/SimpleEnvironmentVariablesProvider.java
```diff
@@ -0,0 +1,87 @@
...
+	/**
+	 * @param environmentProperties a {@link Map} of properties used to override any values
+	 * configured in the current {@link Environment}
+	 */
+	public void setEnvironmentProperties(Map<String, String> environmentProperties) {
+		this.environmentProperties = environmentProperties;
+	}
+
+	@Override
+	public Map<String, String> getEnvironmentVariables(ExecutionContext executionContext) {
```
cppwfs: Why are we passing in a executionContext when it is not used.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/NoOpEnvironmentVariablesProvider.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * A simple no-op implementation of the {@link EnvironmentVariablesProvider}.  It returns
+ * an empty {@link Map}.
+ *
+ * @author Michael Minella
+ *
+ * @since 1.0.2
+ */
+public class NoOpEnvironmentVariablesProvider implements EnvironmentVariablesProvider {
```
cppwfs: Do we need the NoOp impl?  Since it isn't being used?

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/SimpleEnvironmentVariablesProvider.java
```diff
@@ -0,0 +1,87 @@
...
+
+		return environmentProperties;
+	}
+
+	private Map<String, String> getCurrentEnvironmentProperties() {
+		Map<String, String> currentEnvironment = new HashMap<>();
+
+		Set<String> keys = new HashSet<>();
+
+		for (PropertySource<?> propertySource : ((AbstractEnvironment) this.environment).getPropertySources()) {
```
jvalkeal: I'd add null check for `environment` as it's null for default and there is no constructor asserting that it's set. It'd be handy to simply skip environment if it's not set.

jvalkeal:
I'd like to see `SimpleEnvironmentVariablesProvider` to work for simple use case where I can pass in a simple map and not use `environment` if it's not set. Now I had to come up with an impl like:
```
final Map<String, String> environmentProperties = new HashMap<>();
environmentProperties.put("spring.profiles.active", "worker");
partitionHandler.setEnvironmentVariablesProvider(new EnvironmentVariablesProvider() {			
  @Override
  public Map<String, String> getEnvironmentVariables(ExecutionContext executionContext) {
    return environmentProperties;
  }
});
```
I'd like to just use(for passing map and not use environment):
```
new SimpleEnvironmentVariablesProvider(environmentProperties);
```

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/SimpleEnvironmentVariablesProvider.java
mminella: The interface requires it.  The reason the interface requires it is because that's the unique piece of data between each partition and if the environment needs to be unique for each partition, this is what you'd use to know how.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/NoOpEnvironmentVariablesProvider.java
mminella: I can remove it.  I thought it would be useful for some, but no worries.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/SimpleEnvironmentVariablesProvider.java
mminella: Agreed on the null check.  As to not setting it, I can't do that in this impl because I need it to do the exact same thing that the current version does (to prevent it from being a breaking change).

mminella:
In order for this to not be a breaking change, I have the `SimpleEnvironmentVariablesProvider` doing the exact same thing the old code did.  I realize that it requires a custom impl for you with YARN, but as I emphasized before, I expect this impl to be useful for most other could impls.

cppwfs:
LGTM Merged

Issue #179, opened 2016-07-13
Tasks terminated by external signals are not recorded


mminella:
If a task that is executing is terminated by an external signal, we should be able to catch that (on graceful shutdowns) and record the end of the task.  An example of this is PCF canceling an executing task.

PR #184, opened 2016-07-25
Added code to record task completion when process is terminated grace…

mminella:
…fully

Resolves spring-cloud/spring-cloud-task#179

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
@@ -79,6 +79,8 @@ public class TaskLifecycleListener implements ApplicationListener<ApplicationEve
 
 	private boolean started = false;
 
+	private boolean finished = false;
+
 	private TaskNameResolver taskNameResolver;
 
 	private ApplicationArguments applicationArguments;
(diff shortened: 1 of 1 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -266,4 +266,9 @@ public boolean isRunning() {
 	public int getPhase() {
 		return 0;
 	}
+
+	@Override
+	public void destroy() throws Exception {
```
cppwfs: This will call the doTaskEnd twice at the end of a task that ends normally.  Once for the onApplicationEvent then again for the destroy.  
Do we want to put a flag here that will prevent it from calling the destroy if we have already recorded our results?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -167,8 +170,10 @@ else if(this.applicationFailedEvent != null){
 			if(this.closeContext && this.context.isActive()) {
```
cppwfs: We need to move the `this.finished = true;` to above where we close the context.  Else the onDestroy fires prior  to the finished being set to true.  I can do this on merge if you like.

mminella: Agreed.

cppwfs:
Did the cleanup as discussed.  Rebased Merged

PR #185, opened 2016-08-06
Fix NPE when return an eventExitCode before ApplicationFailedEvent

72MiguelGomes:
**Problem:**

NPE when an eventExitCode is fired before ApplicationFailedEvent.

**Example:**

```
public class HelloCommandListRunner implements CommandLineRunner, ExitCodeExceptionMapper {

    @Override
    public void run(String... args) throws Exception {
        throw new ExceptionTest("test");
    }

    @Override
    public int getExitCode(Throwable exception) {

        if (exception.getCause() instanceof ExceptionTest) {
            return 10;
        }

        return 0;
    }
}

```

This causes a NPE because the error code is different from 0 but _this.applicationFailedEvent_ is not set.

```
if(this.taskExecution.getExitCode() != 0){
				taskExecution.setExitMessage(invokeOnTaskError(taskExecution,
						this.applicationFailedEvent.getException()).getExitMessage());
			}

```

I also remove the _doTaskEnd();_ for _ExitCodeEvent_ because every time we sent an ExitCodeEvent we also send an _ApplicationFailedEvent_, so the second one will end the task, otherwise the information from the last event (_ApplicationFailedEvent_) will be never recorded.

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
@@ -125,7 +125,6 @@ public class TaskLifecycleListener implements ApplicationListener<ApplicationEve
 		}
 		else if(applicationEvent instanceof ExitCodeEvent){
 			this.exitCodeEvent = (ExitCodeEvent) applicationEvent;
-			doTaskEnd();
 		}
 		else if(applicationEvent instanceof ApplicationReadyEvent) {
 			doTaskEnd();
@@ -159,7 +158,7 @@ public class TaskLifecycleListener implements ApplicationListener<ApplicationEve
 				this.taskExecution.setExitMessage(stackTraceToString(this.applicationFailedEvent.getException()));
 			}
 
-			if(this.taskExecution.getExitCode() != 0){
+			if (this.applicationFailedEvent != null && this.taskExecution.getExitCode() != 0) {
 				taskExecution.setExitMessage(invokeOnTaskError(taskExecution,
 						this.applicationFailedEvent.getException()).getExitMessage());
 			}
--- a/spring-cloud-task-core/src/test/java/org/springframework/cloud/task/listener/TaskLifecycleListenerTests.java
+++ b/spring-cloud-task-core/src/test/java/org/springframework/cloud/task/listener/TaskLifecycleListenerTests.java
@@ -27,6 +27,7 @@ import org.junit.Before;
 import org.junit.Test;
 
 import org.springframework.boot.ApplicationArguments;
+import org.springframework.boot.ExitCodeEvent;
 import org.springframework.boot.SpringApplication;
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@72MiguelGomes Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-task&pullRequestId=185)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-task&pullRequestId=185) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@72MiguelGomes Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-task&pullRequestId=185)!

mminella:
Unless I'm missing something, I don't see anywhere in the Spring Boot code that guarantees that when an `ExitCodeEvent` is thrown that either a `ApplicationFailedEvent` or `ApplicationReadyEvent` are thrown.  If I'm correct, then we'd need the `doEndTask` that was removed.  And in either case, the `doEndTask` should only actually execute once based on the finished flag.  cc: @philwebb

philwebb:
@mminella I think that `ApplicationFailedEvent` and `ApplicationReadyEvent` are fired based on the `run` method and `ExitCodeEvent` is fired only on a non zero `exitCode`. It probably depends on how `SpringApplication.exit` gets called as to of the other events get fired.

I think it's legitimate to say that an application started correctly (so an `ApplicationReadyEvent` is fired) and then exits (so an `ExitCodeEvent` is fired). In other words I'm not sure an `Exit` event necessarily means a failure. Let me know if we need to change something in Boot or if more documentation is needed.

mminella:
@philwebb What we really need to understand is that we can expect either a `ApplicationFailedEvent` or an `ApplicationReadyEvent` after an `ExitCodeEvent` has been sent.  I'm unclear on what the contract of these are.  Part of this PR is removing a call to our `doEndTask` method when an `ExitCodeEvent` is received which means that we'd rely on the other two events exclusively.

philwebb:
The `ApplicationFailedEvent` and `ApplicationReadyEvent` are unrelated to `ExitCodeEvent`. You can expect `ApplicationFailedEvent` or `ApplicationReadyEvent` to be fired when `run` is called. It's still possible I think to get these events _before_ an `ExitCodeEvent` since it's possible that `run` returns then the user call `SpringApplication.exit`.

72MiguelGomes:
@mminella You are right in part but if you look at the code just the ApplicationFailedEvent has information about exception that was thrown. and if you close the context when you get an ExitCodeEvent you will not have information about it.

mminella:
Thanks for the efforts on this one.  Merged as 9ccbe60cbaadb8f1950e6f31fd57e12117e4f9a5

PR #198, opened 2016-08-25
Truncates the exit and error messages to 2500 characters

cppwfs:


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -55,6 +58,7 @@
  */
 public class JdbcTaskExecutionDao implements TaskExecutionDao {
 
+	public static int MAX_MESSAGE_SIZE = 2500;
```
mminella: This needs to be configurable.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/SimpleTaskRepository.java
```diff
@@ -36,6 +36,7 @@
 
 	public static final int MAX_EXIT_MESSAGE_SIZE = 2500;
 	public static final int MAX_TASK_NAME_SIZE = 100;
+	public static final int MAX_ERROR_MESSAGE_SIZE = 2500;
```
mminella: Am I missing something or is this still not configurable?

cppwfs: My recommendation here is for the user to create their own TaskRepository.  Else they will have to set these via properties. And that didn't seem that clean to me.  So argue with me if you disagree.

cppwfs: After our discussion.  they will be configurable.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/SimpleTaskRepository.java
```diff
@@ -45,12 +46,27 @@
...
+	private int maxErrorMessageSize = MAX_ERROR_MESSAGE_SIZE;
+
 	public SimpleTaskRepository(FactoryBean<TaskExecutionDao> taskExecutionDaoFactoryBean){
 		Assert.notNull(taskExecutionDaoFactoryBean, "A FactoryBean that provides a TaskExecutionDao is required");
 
 		this.taskExecutionDaoFactoryBean = taskExecutionDaoFactoryBean;
 	}
 
+	public SimpleTaskRepository(FactoryBean<TaskExecutionDao> taskExecutionDaoFactoryBean, int maxExitMessageSize,
+			int maxTaskNameSize, int maxErrorMessageSize){
```
mminella: Thoughts about `Integer` instead of `int` so that we don't need to configure them all explicitly?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/SimpleTaskRepository.java
```diff
@@ -126,12 +143,36 @@ private void validateExitInformation(long executionId, Integer exitCode,  Date e
...
+		this.maxTaskNameSize = maxTaskNameSize;
+	}
+
+	public int getMaxErrorMessageSize() {
+		return maxErrorMessageSize;
+	}
+
+	public void setMaxErrorMessageSize(int maxErrorMessageSize) {
+		this.maxErrorMessageSize = maxErrorMessageSize;
+	}
```
mminella: Why the three getters?

mminella:
LGTM.  Merged as 4e7e75cc82e45a6bd92477732c7d0439fdae8865

Issue #177, opened 2016-07-13
`spring-cloud-task-dependencies` contains non-spring cloud task dependency management


snicoll:
The public BOM must only contain dependency management for the current project and not force any version for "external" dependencies. As it stands, `spring-batch-core` is probably the most annoying but spring cloud stream is also affected (as its own bom provides the version to use). 

There is a pattern to deal with this in Spring Cloud and you are probably using it already so it's a matter of moving those dependencies in the other pom.

mminella:
@snicoll What other Spring Cloud projects can you provide as an example of this?  And what do you mean by "the other pom"?

snicoll:
The spring cloud team put that infra in place and spent some time to get it right so I'd ask them.

mbogoevici:
The non SCSt dependencies in the SCSt-dependencies BOM have to go. We've had them there for a while and keeping them there was an oversights.

snicoll:
The `spring-cloud-task` bom is still wrong. Can we please fix it? ping @dsyer

dsyer:
Bump. It seems like 1.0.2 was released and this wasn't addressed.

PR #200, opened 2016-08-29
Refactored dependencies to prevent incorrect versions from being brought in

mminella:
This PR removes all references to version numbers for projects other than Spring Cloud Task projects from the `spring-cloud-task-dependencies` POM.  This allows for the independent management of those dependencies.

Resolves spring-cloud/spring-cloud-task#177

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,61 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-deployer-spi</artifactId>
+				<version>${spring-cloud-deployer-spi.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-deployer-local</artifactId>
+				<version>${spring-cloud-deployer-local.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-deployer-resource-support</artifactId>
+				<version>${spring-cloud-deployer-resource-support.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.batch</groupId>
+				<artifactId>spring-batch-core</artifactId>
+				<version>${spring-batch.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.batch</groupId>
+				<artifactId>spring-batch-infrastructure</artifactId>
+				<version>${spring-batch.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.batch</groupId>
+				<artifactId>spring-batch-integration</artifactId>
+				<version>${spring-batch.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-deployer-resource-maven</artifactId>
+				<version>${spring-cloud-deployer-resource-maven.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-stream-test-support-internal</artifactId>
+				<version>${spring-cloud-stream-binder-rabbit.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-starter-stream-rabbit</artifactId>
+				<version>${spring-cloud-stream-binder-rabbit.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-stream</artifactId>
+				<version>${spring-cloud-stream.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-stream-test-support</artifactId>
+				<version>${spring-cloud-stream.version}</version>
+			</dependency>
 		</dependencies>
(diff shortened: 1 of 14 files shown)


snicoll:
So that's not good, sorry. Dependencies version must be located in a `dependencyManagement` section in isolated places. Child modules must only specify `groupId` and `artifactId` and never have a `version` tag.

Spring Cloud has a special structure with two POMs to address this: one (the BOM) that is used to provide dependency management for the local modules and one pom to add dependency management for _dependencies_ (3rd party libs, boot, other cloud projects, etc).

I've asked @dsyer to provide you an example of that. Maybe you could have a look to other projects in Spring Cloud to see how they're doing this. One way to get started is to look at one bom that is imported in `spring-cloud-dependencies` and see how it's being used within its own project?

mminella:
@snicoll @dsyer pointed me to `spring-cloud-stream` but I'm not seeing what you're describing there.  What, specifically, should I be verifying?  Is there a maven command that you're running that shows you something is amiss that I should be running to validate my setup?  Are we sure that `spring-cloud-stream` is also right (since that's where I'm begin pointed)?

pom.xml
```diff
@@ -61,6 +61,61 @@
...
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-deployer-resource-maven</artifactId>
+				<version>${spring-cloud-deployer-resource-maven.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-stream-test-support-internal</artifactId>
+				<version>${spring-cloud-stream-binder-rabbit.version}</version>
```
cppwfs: Should this version be based on `spring-cloud-stream.version`

pom.xml
```diff
@@ -76,6 +131,14 @@
 	</modules>
 
 	<properties>
+		<spring-cloud-stream.version>1.0.2.RELEASE</spring-cloud-stream.version>
+		<spring-cloud-deployer-spi.version>1.0.2.RELEASE</spring-cloud-deployer-spi.version>
+		<spring-cloud-deployer-local.version>1.0.2.RELEASE</spring-cloud-deployer-local.version>
+		<spring-cloud-stream-binder-rabbit.version>1.0.2.RELEASE</spring-cloud-stream-binder-rabbit.version>
+		<spring-cloud-deployer-resource-support.version>1.0.2.RELEASE</spring-cloud-deployer-resource-support.version>
+		<spring-cloud-deployer-resource-maven.version>1.0.2.RELEASE</spring-cloud-deployer-resource-maven.version>
```
cppwfs: Is this being used?  I see that it is commented out in the spring-cloud-task-core/pom.xml

spring-cloud-task-core/pom.xml
```diff
@@ -27,6 +27,7 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-deployer-resource-maven</artifactId>
 			<optional>true</optional>
+			<!--<version>${spring-cloud-deployer-resource-maven.version}</version>-->
```
cppwfs: Do we need this anymore?

pom.xml
```diff
@@ -118,6 +181,7 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-source-plugin</artifactId>
+				<version>3.0.1</version>
```
cppwfs: Just curious.  Why are we setting this version?

spring-cloud-task-samples/task-events/pom.xml
```diff
@@ -14,6 +14,7 @@
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
 		<version>1.4.0.RELEASE</version>
+		<relativePath/>
```
cppwfs: nice

spring-cloud-task-samples/batch-events/pom.xml
```diff
@@ -50,11 +51,13 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-starter-stream-rabbit</artifactId>
+			<version>1.0.2.RELEASE</version>
```
cppwfs: With the addition of adding specific version numbers to our samples, it makes me wonder if we should move samples into their own repo.

pom.xml
mminella: Good catch.

pom.xml
mminella: Clears up a warning from Maven.

spring-cloud-task-samples/batch-events/pom.xml
mminella: Something to talk about but definitely not something we want to do for this PR...

pom.xml
mminella: It's used later in this file (line 97).

spring-cloud-task-core/pom.xml
mminella: Nope.

spring-cloud-task-samples/batch-events/pom.xml
cppwfs: Agreed!

cppwfs:
LGTM

cppwfs:
Rebased, squashed merged.

snicoll:
@cppwfs where is the final commit? #177 is still open.

cppwfs:
@snicoll here we go 
For Master:  https://github.com/spring-cloud/spring-cloud-task/commit/c14d2efca3a9b70e4aa397145a020daf86567c6c

For 1.0.x: https://github.com/spring-cloud/spring-cloud-task/commit/5d08953b799ea3ca900a3418c4c6804ae7c9c4e7

PR #215, opened 2016-09-22
Updated task name so that it will work with kubernetes out of the box.

### LINKED ISSUE #210: Update Partition Sample App Name
The application name should run on k8's without modification ie. replace `PartitionBatchJobTask` with `partitionedbatchjobtask`

cppwfs:
resolves #210

### DIFF
--- a/spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
+++ b/spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
@@ -93,7 +93,7 @@ public class JobConfiguration {
 
 	@Bean
 	public PartitionHandler partitionHandler(TaskLauncher taskLauncher, JobExplorer jobExplorer) throws Exception {
-		Resource resource = resourceLoader.getResource("maven://io.spring.cloud:partitioned-batch-job:1.1.0.M1");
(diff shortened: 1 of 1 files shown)


spring-cloud-task-samples/partitioned-batch-job/src/main/java/io/spring/JobConfiguration.java
```diff
@@ -104,7 +104,7 @@ public PartitionHandler partitionHandler(TaskLauncher taskLauncher, JobExplorer
 		partitionHandler.setCommandLineArgsProvider(new PassThroughCommandLineArgsProvider(commandLineArgs));
 		partitionHandler.setEnvironmentVariablesProvider(new SimpleEnvironmentVariablesProvider(this.environment));
 		partitionHandler.setMaxWorkers(1);
-		partitionHandler.setApplicationName("PartitionedBatchJobTask");
```
trisberg: I added a `toLowerCase()` for the id generated in the k8s deployer, so if you prefer, you can keep the name as it was

cppwfs:
I thought the deployer would throw an exception if it was not an properly formatted.

trisberg:
Since it's a generated id it's not necessary to throw an exception, we can try to make it fit - https://github.com/trisberg/spring-cloud-deployer-kubernetes/blob/ce7b8d8e0a64ebfe4aee373ca29accfe903e95d6/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesTaskLauncher.java#L114-L121

cppwfs:
Gotcha.  Closing PR as complete.

PR #220, opened 2016-10-05
Adds support for externalExecutionId

### LINKED ISSUE #80: User wants to record externalExecutionID for each task execution
Allows a user to pass in a externalExecutionID (DepoymentID) if one is generated by the system (CF, Yarn, etc) that is executing the task.  This externalExecutionID will be stored in the taskExecutionTable.  
possibly --spring.cloud.task.externalexecutionid

Depends on #115

cppwfs:
resolves #80


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/TaskExecution.java
```diff
@@ -61,6 +62,11 @@
 	private String exitMessage;
 
 	/**
+	 * id assigned to the task by the platform.
```
cppwfs: capitalize Id

cppwfs: Capitalize Id

spring-cloud-task-integration-tests/pom.xml
```diff
@@ -119,7 +119,7 @@
 						<artifactId>maven-surefire-plugin</artifactId>
 						<version>2.12.4</version>
 						<configuration>
```
cppwfs: set this back to true

spring-cloud-task-samples/timestamp/pom.xml
```diff
@@ -61,6 +61,10 @@
 			<groupId>com.h2database</groupId>
 			<artifactId>h2</artifactId>
 		</dependency>
+		<dependency>
```
cppwfs: Remove MariaDB dependency

cppwfs:
Need to add documentation as well.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -187,9 +198,18 @@ private void doTaskStart() {
...
+			if(this.taskExecutionId != null) {
+				TaskExecution taskExecution = taskExplorer.getTaskExecution(this.taskExecutionId);
+				Assert.notNull(taskExecution, String.format("Invalid TaskExecution, ID %s not found", this.taskExecutionId));
+				Assert.isNull(taskExecution.getEndTime(), String.format(
+						"Invalid TaskExecution, ID %s task is already complete", this.taskExecutionId));
+				this.taskExecution = this.taskRepository.startTaskExecution(this.taskExecutionId,
+						this.taskNameResolver.getTaskName(), new Date(), args, externalExecutionId);
+			}
+			else {
+				this.taskExecution = this.taskRepository.createTaskExecution(
```
mminella: In 2.0, we should consider passing a `TaskExecution` instead of the itemized parameters to minimize breaking changes like this going forward.  We can always document what we do with it internally.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/SimpleTaskRepository.java
```diff
@@ -100,15 +101,34 @@ public TaskExecution completeTaskExecution(long executionId, Integer exitCode, D
...
+				taskExecutionDao.createTaskExecution(taskName, startTime, arguments, externalExecutionId);
 		logger.debug("Creating: " + taskExecution.toString());
 		return taskExecution;
 	}
 
+	@Override
+	public TaskExecution createTaskExecution() {
+		initialize();
+		TaskExecution taskExecution =
+				taskExecutionDao.createTaskExecution(null, null, new ArrayList<String>(0), null);
```
mminella: `Collections.emptyList`.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -92,18 +95,26 @@
 	@Value("${spring.cloud.task.closecontext.enable:true}")
 	private Boolean closeContext;
 
+	@Value("${spring.cloud.task.executionid:}")
+	private Integer taskExecutionId;
+
+	@Value("${spring.cloud.task.external.executionid:#{null}}")
```
mminella: This really should be `spring.cloud.task.external-execution-id` (no extra `.` and use `-` to separate multi word like boot).

mminella: There are also no tests that use this property.  We need to test the injection works.  Also, is the `null` really needed?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
cppwfs: Agreed
https://github.com/spring-cloud/spring-cloud-task/issues/223

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/SimpleTaskRepository.java
cppwfs: good catch

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
cppwfs: Without the `null` default it places an empty string for the externalExecutionId.  This would be inconsistent with the defaults for errorMessage and exitMessage which are set to null

cppwfs:
Latest commit: Squashed, rebased to master and updated based on code review.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
mminella: Why does it fail in the shell?  That is the standard syntax for boot properties?  See here: http://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html

mminella:
Merged as 17e42505869719b1ca3de0efe5d2b4705dece8ec

PR #227, opened 2016-10-21
Adding a few more tests.

cppwfs:
Added unit tests to classes that had no coverage.
Added unit tests to classes bump up coverage.

### DIFF
--- a/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/listener/TaskBatchExecutionListenerTests.java
+++ b/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/listener/TaskBatchExecutionListenerTests.java
@@ -95,18 +95,7 @@ public class TaskBatchExecutionListenerTests {
 		assertEquals(1, jobExecutionIds.size());
 		assertEquals(1, taskExplorer.getTaskExecution(jobExecutionIds.iterator().next()).getExecutionId());
 	}
-
-	@Test
-	public void testBeanFactoryReturnsSameInstance() throws Exception{
-		this.applicationContext = SpringApplication.run(new Object[] {JobConfiguration.class,
-				PropertyPlaceholderAutoConfiguration.class,
-				EmbeddedDataSourceConfiguration.class,
-				BatchAutoConfiguration.class,
-				TaskBatchAutoConfiguration.class}, ARGS);
-
-		TaskBatchExecutionListenerFactoryBean bean = this.applicationContext.getBean(TaskBatchExecutionListenerFactoryBean.class);
-		assertTrue(bean.getObject() == bean.getObject());
-	}
+	
 	@Test
 	public void testMultipleDataSources() {
 		this.applicationContext = SpringApplication.run(new Object[] {JobConfigurationMultipleDataSources.class,
--- a/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/listener/TaskBatchExecutionListenerTests.java
+++ b/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/listener/TaskBatchExecutionListenerTests.java
@@ -60,6 +60,7 @@ import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseBuilder;
 import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseType;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 /**
  * @author Michael Minella
@@ -95,6 +96,17 @@ public class TaskBatchExecutionListenerTests {
 		assertEquals(1, taskExplorer.getTaskExecution(jobExecutionIds.iterator().next()).getExecutionId());
 	}
 
+	@Test
+	public void testBeanFactoryReturnsSameInstance() throws Exception{
+		this.applicationContext = SpringApplication.run(new Object[] {JobConfiguration.class,
+				PropertyPlaceholderAutoConfiguration.class,
+				EmbeddedDataSourceConfiguration.class,
+				BatchAutoConfiguration.class,
+				TaskBatchAutoConfiguration.class}, ARGS);
+
+		TaskBatchExecutionListenerFactoryBean bean = this.applicationContext.getBean(TaskBatchExecutionListenerFactoryBean.class);
+		assertTrue(bean.getObject() == bean.getObject());
+	}
 	@Test
 	public void testMultipleDataSources() {
 		this.applicationContext = SpringApplication.run(new Object[] {JobConfigurationMultipleDataSources.class,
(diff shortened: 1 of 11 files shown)


spring-cloud-task-core/src/test/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfigurationTests.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = {ResourceLoadingAutoConfiguration.class})
+public class ResourceLoadingAutoConfigurationTests {
+
+	@Autowired
+	ApplicationContext context;
+
+	@Test
+	public void testTaskLauncher() {
```
mminella: It would be good to test the conditionals as well in this test.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDaoTests.java
```diff
@@ -105,4 +110,31 @@ public void completeTaskExecution(){
 				taskExecutionMap.get(expectedTaskExecution.getExecutionId()));
 	}
 
+	@Test
+	public void testJobQueries(){
+		List<TaskExecution> expectedTaskExecutionList = new ArrayList<TaskExecution>(2);
```
mminella: Just an FYI, since we compile to 1.7, we can use diamond operators `<>`.

spring-cloud-task-stream/src/test/java/org/springframework/cloud/task/batch/listener/JobInstanceEventTests.java
```diff
@@ -36,4 +37,23 @@ public void testConstructor() {
...
 		assertEquals(JOB_NAME, jobInstanceEvent.getJobName());
 	}
+
+	@Test
+	public void testEmptyConstructor() {
+		JobInstanceEvent jobInstanceEvent = new JobInstanceEvent();
+		assertNull(jobInstanceEvent.getJobName());
+	}
+
+	@Test(expected = NullPointerException.class)
```
mminella: I'm 99% sure we should never *expect* a NPE.  If this test passes, I feel like there is something wrong...

spring-cloud-task-stream/src/test/java/org/springframework/cloud/task/launcher/TaskLaunchConfigurationTests.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * @author Glenn Renfro
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = {TaskLauncherConfiguration.class})
+public class TaskLaunchConfigurationTests {
```
mminella: If we're going to test this, we should test all the cases (including the ones where ours shouldn't show up, etc.

spring-cloud-task-stream/src/test/java/org/springframework/cloud/task/launcher/TaskLaunchRequestTests.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * @author Glenn Renfro
+ */
+public class TaskLaunchRequestTests {
+	public static final String URI = "http://myURI";
+
+	@Test
+	public void testEquals() {
+		List args = new ArrayList<String>();
+		Map map = new HashMap<String,String>();
```
mminella: Generics for both the `List` and `Map`?

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/listener/TaskBatchExecutionListenerTests.java
```diff
@@ -91,6 +97,17 @@ public void testAutobuiltDataSource() {
 	}
 
 	@Test
+	public void testBeanFactoryReturnsSameInstance() throws Exception{
```
cppwfs: creates a `TaskBatchExecutionListener` instance and it is returned.  The second getObject call return the previously created instance.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDaoTests.java
cppwfs: DOH!

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfigurationTests.java
cppwfs: I'm wondering if the second conditional is even necessary.  In that the dependency is already present, would there be a case that it wouldn't.

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/listener/TaskBatchExecutionListenerTests.java
mminella: Yeah...but isn't that all functions of Spring Core/Bean Factories and not SCT?

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfigurationTests.java
mminella: Which "second condition"?

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/listener/TaskBatchExecutionListenerTests.java
cppwfs: The getObject method of the TaskBatchExecutionListenerFactoryBean in the SCT project.  And it was marked red as not covered by code coverage.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/configuration/ResourceLoadingAutoConfigurationTests.java
cppwfs: The case where ConditionalOnMissingClass for MavenResourceLoader is true.  Not sure if this case will occur.

mminella: The `MavenResourceLoader` is not required within an app so it's better to be safe than sorry.

cppwfs:
It does test that the instance cached.  
But it will be removed.

PR #239, opened 2016-11-02
Add Application Name to TaskLaunchRequest for launched task

### LINKED ISSUE #238: TaskLaunchRequest should allow user to set the app name
Developers wants the app name for a task that is launched by the TaskSInk to be obtained from the TaskLaunchRequest.  
If app name is not set then it will use the default of `Task-<hashCode of TaskLaunchRequest>`

cppwfs:
resolves #238

### DIFF
--- a/spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLaunchRequest.java
+++ b/spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLaunchRequest.java
@@ -23,8 +23,10 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * Request that contains the maven repository and property information required by the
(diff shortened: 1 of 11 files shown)


spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLaunchRequest.java
```diff
@@ -90,6 +96,26 @@ public String getUri() {
...
+	}
+
+	/**
+	 * Sets the name to be applied to the launched task.   If set
+	 * 	to null then the launched task name will be "Task-<hash code of the task
+	 * 	launch request>.
+	 * @param applicationName the name to be
+	 */
+	public void setApplicationName(String applicationName) {
+		this.applicationName = applicationName == null ? "Task-" + hashCode() :
```
mminella: Why are we using hash code here?  And it doesn't feel right to me to change the state of the object with the hash code like that.  Correct me if I'm wrong, but after calling this method, if the name is foo-123 (123 is my hash code), if I call `hashCode()` on this object now, I won't get 123 again.  If that's the case, we shouldn't advertise this as the object's hash code.  I'd take a random UUID personally...

mminella: Also, we should use `StringUtils#hasText` instead of just a null check.

mminella:
Closed as 4a1a2c3c88435233d55fef276ee7498cc35bb481

PR #257, opened 2016-12-02
TaskBatchAutoConfiguration now uses datasource from TaskConfigurer

### LINKED ISSUE #252: Get datasource from `TaskConfigurer` when building batch listener
When a user defines multiple `DataSource` beans in their context, they need to create a `TaskConfigurer` to identify which one to use.  However, the autoconfiguration for the `TaskBatchExecutionListener` doesn't use that reference (although it should).  Because of this, the user needs to override the configuration of the `TaskBeanExecutionListener` in this case.  We can make this better by just recycling the reference from the `TaskConfigurer`.

cppwfs:
resolves #252


mminella:
LGTM.  Merged as 1185421803055ce9c9327ac755ec37a2fab6c267

PR #268, opened 2017-01-11
Change deployer arguments into environment variables

ssiahetiong:
Since chronos appends the arguments to the mesos task id, the length of task id is exceeding its max length. Moving the default args to environment variables shortens the task id.


mminella:
There are implications to just moving those arguments from command line args to environment variables.  Things like:
* They are not passed as job parameters to a job
* They are not recorded by the task execution

Because of that, I'd want to make it configurable so that you can choose and have it default to command line args.  If that is ok, please update the PR accordingly.

mminella:
Thanks for the PR.  I took a slightly different approach and committed it but the basic functionality you are looking for should now be available.

PR #273, opened 2017-01-16
Do not update the externalExecutionId if null when starting TaskExecu…

### LINKED ISSUE #271: Do not update the externalExecutionId field if null when executing startTaskExecution

cppwfs:
…tion

resolves #271

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
@@ -196,19 +196,22 @@ public class JdbcTaskExecutionDao implements TaskExecutionDao {
 			String externalExecutionId, Long parentExecutionId) {
 		TaskExecution taskExecution = new TaskExecution(executionId, null, taskName,
 				startTime, null, null, arguments,null, externalExecutionId, parentExecutionId);
-		Object[] queryParameters = new Object[]{ startTime, taskName,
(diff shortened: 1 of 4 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -165,16 +169,29 @@ public TaskExecution createTaskExecution(String taskName,
...
+	public TaskExecution startTaskExecution(long executionId, String taskName,
+			Date startTime, List<String> arguments,
 			String externalExecutionId) {
-		TaskExecution taskExecution = new TaskExecution(executionId, null, taskName,
-				startTime, null, null, arguments, null, externalExecutionId);
+		TaskExecution taskExecution = new TaskExecution(executionId, null,
+				taskName, startTime, null, null, arguments, null,
+				externalExecutionId);
+
+		Object[] queryParameters = new Object[]{ startTime, taskName,
```
mminella: Very small nit pick here.  For efficiency, it would be better to create this array in an `else` of the `if` statement below (instead of always creating two arrays.  Same goes for the `argTypes` arrays as well.

cppwfs: I can make the change.  I originally had it in the else.   CTRL-Z baby!

mminella:
Merged as a7fe76c7d6c286599c0cd6b331a1f2bd39f5fe75

PR #285, opened 2017-02-27
Add Note to TaskConfigurer Ref Doc

### LINKED ISSUE #272: Update Reference docs to include note on transaction managers
It would be a good idea to include a note on the : 
 http://docs.spring.io/spring-cloud-task/current-SNAPSHOT/reference/htmlsingle/#features-task-configurer
on how creating your own TaskConfigurer would be a solution when using a transaction manager that is not the ones provided by the DefaultTaskConfigurer.  For example a common transaction manager such as `JpaTransactionManager` .

cppwfs:
Reminds users to create their own TaskConfigurer implementation when the DefaultTaskConfigurer does not meet their needs.

resolves #272

### DIFF
--- a/spring-cloud-task-docs/src/main/asciidoc/features.adoc
+++ b/spring-cloud-task-docs/src/main/asciidoc/features.adoc
@@ -236,6 +236,12 @@ repository) to be used.
 `ResourcelessTransactionManager` if it is not.
 |===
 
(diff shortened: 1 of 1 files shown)


mminella:
Merged as 901b21cbe833ee2271c4e2e2e07ab94f3af85da8

PR #288, opened 2017-03-02
Updated the migration scripts to use all caps for table names

### LINKED ISSUE #287: Error upgrading MySQL database
I'm getting an error running the 1.2 migration script on a MySQL database:

```
alter table task_execution add PARENT_EXECUTION_ID BIGINT;
```

results in: Code: 1146 SQL State: 42S02 --- Table 'test.task_execution' doesn't exist

The table name should be uppercase since that is how it is defined in `schema-mysql.sql`

cppwfs:
In the original migration scripts the table names were all lower case.  However the original table definitions were capitalized, and on file systems that are case sensitive,
mysql fails to update the table.  Because of the 2 different cases.

resolves #287


trisberg:
I think we should do the same for 1.1.x migration scripts as well?

cppwfs:
on it

trisberg:
LGTM, except I just noticed we don't have a 1.2.x migration script for db2, we should add that since schema-db2.sql script was added for 1.1.0

PR #290, opened 2017-03-03
Updated samples to use latest boot

### LINKED ISSUE #289: Update task samples to boot 1.5.2
Also update tests to use updated stream paradigms as well.

cppwfs:
Update integration tests to use latest stream paradigm
Update tests to be in compliance with 1.5.2.

resolves #289

### DIFF
--- a/spring-cloud-task-integration-tests/src/test/java/org/springframework/cloud/task/listener/BatchExecutionEventTests.java
+++ b/spring-cloud-task-integration-tests/src/test/java/org/springframework/cloud/task/listener/BatchExecutionEventTests.java
@@ -27,6 +27,7 @@ import org.junit.After;
 import org.junit.ClassRule;
 import org.junit.Test;
 
+import org.springframework.batch.core.scope.context.ChunkContext;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.cloud.stream.annotation.EnableBinding;
@@ -189,7 +190,7 @@ public class BatchExecutionEventTests {
 	public static class ItemProcessListenerBinding {
 
 		@StreamListener(Sink.INPUT)
-		public void receive(Object object) {
+		public void receive(String object) {
 			itemProcessLatch.countDown();
 		}
 	}
@@ -200,7 +201,7 @@ public class BatchExecutionEventTests {
 	public static class ChunkEventsListenerBinding {
 
 		@StreamListener(Sink.INPUT)
-		public void receive(Object chunkContext) {
+		public void receive(ChunkContext chunkContext) {
 			chunkEventsLatch.countDown();
 		}
 	}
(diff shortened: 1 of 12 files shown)


spring-cloud-task-integration-tests/src/test/java/org/springframework/cloud/task/listener/BatchExecutionEventTests.java
```diff
@@ -240,7 +241,7 @@ public void receive(Object exceptionMessage) {
 	public static class ItemWriteEventsListenerBinding {
 
 		@StreamListener(Sink.INPUT)
-		public void receive(Object itemWrite) {
+		public void receive(String itemWrite) {
 			assertTrue("Message should start with '3 items'", itemWrite.toString().startsWith("3 items "));
```
mminella: @mbogoevici Is the requirement to make these changes expected?  No objections, just want to be sure that this isn't an unexpected side effect from something.

mbogoevici: @mminella Does not seem like a side effect - basically the argument was always a String, so now the method is just treating it as such (StreamListener helps with the coercion in certain cases). Doesn't seem to make a difference when the argument is not actually being used.

mminella: @mbogoevici What I'm specifically asking is that, before `Object` worked.  Now we're being required to specify the type (the test fails without this change).

mbogoevici: That's a little odd then. Maybe we can take a look.

mbogoevici: @mminella The test failures were called by a regression - see https://github.com/spring-cloud/spring-cloud-stream/issues/844. 

On the other hand, I would recommend keeping these changes, as they are closer to what a user would be expected to do anyway (i.e. not use Object and downcast, but use the actual expected type as argument).

mminella:
This PR isn't building for me:
```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /Users/mminella/Documents/IntelliJWorkspace/spring-cloud-task/spring-cloud-task-stream/src/test/java/org/springframework/cloud/task/launcher/configuration/TaskConfiguration.java:[40,23] org.springframework.cloud.task.launcher.configuration.TaskConfiguration.TestTaskLauncher is not abstract and does not override abstract method environmentInfo() in org.springframework.cloud.deployer.spi.task.TaskLauncher
[INFO] 1 error
```

mminella:
Merged as 4e68adf5facdc92faff5384af65846faa298848d

PR #292, opened 2017-03-06
Add Ordered interface to all event listeneres

### LINKED ISSUE #291: Add Ordered Interface to EventEmittingJobExecutionListener and other events
I want to force my AfterJob to be executed before `EventEmittingJobExecutionListener` because i do change some properties in the context and i want these to be emitted . 

Currently `EventEmittingJobExecutionListener` is not ordered so it is being executed before the ordered .

If you agree on this i can do a PR

shahbour:
Add Ordered interface to all event listeneres , this is related to issue #291

Resolves #291


spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingItemProcessListener.java
```diff
@@ -64,4 +66,9 @@ else if (item.equals(result)) {
 	public void onProcessError(Object item, Exception e) {
 		messagePublisher.publishWithThrowableHeader("Exception while item was being processed", e.getMessage());
 	}
+
+	@Override
+	public int getOrder() {
+		return 0;
```
cppwfs: In this case the order should be configurable for each of the event emitting listerners.  This can be done via a Spring Boot property i.e. `spring.cloud.task.events.<batch event listener>.order=<order#>` .  For example:
`spring.cloud.task.events.job-execution.order=<order#>`

cppwfs: Default should be Integer.MIN_VALUE instead of zero.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingItemProcessListener.java
```diff
@@ -33,8 +34,9 @@
  *
  * @author Michael Minella
  * @author Glenn Renfro
+ * @author Ali Shahbour
  */
-public class EventEmittingItemProcessListener implements ItemProcessListener {
+public class EventEmittingItemProcessListener implements ItemProcessListener,Ordered {
```
cppwfs: should have a space following the comma.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingItemReadListener.java
```diff
@@ -34,8 +35,9 @@
  * via the {@link BatchJobHeaders.BATCH_EXCEPTION} message header.
  *
  * @author Glenn Renfro
+ * @author Ali Shahbour
  */
-public class EventEmittingItemReadListener implements ItemReadListener {
+public class EventEmittingItemReadListener implements ItemReadListener,Ordered {
```
cppwfs: add space following comma

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingJobExecutionListener.java
```diff
@@ -27,8 +28,9 @@
  *
  * @author Michael Minella
  * @author Glenn Renfro
+ * @author Ali Shahbour
  */
-public class EventEmittingJobExecutionListener implements JobExecutionListener {
+public class EventEmittingJobExecutionListener implements JobExecutionListener,Ordered {
```
cppwfs: add space following comma

cppwfs:
Need to implement Ordered on a new EventEmittingChunkListener.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingItemProcessListener.java
shahbour: What about using Ordered.HIGHEST_PRECEDENCE

shahbour: can you please elaborate more on how you wish this to be implemented . Should i read them one by one in `JobExecutionListenerConfiguration` as they can't be injected into each listener

shahbour: If the below is OK i will proceed 

```
	@ConfigurationProperties(prefix = "spring.cloud.task.events")
	public class TaskEventsProperties {

		private ListenerProperties jobExecution;

		public ListenerProperties getJobExecution() {
			return jobExecution;
		}

		public void setJobExecution(ListenerProperties jobExecution) {
			this.jobExecution = jobExecution;
		}

		public  class ListenerProperties {

			private int order;

			public int getOrder() {
				return order;
			}

			public void setOrder(int order) {
				this.order = order;
			}
		}
	}
```

another issue the prefix should be `spring.cloud.task.events` or `spring.cloud.task.batch.events`

cppwfs: makes sense

cppwfs: Why do we need ListenerProperties in an inner class?

cppwfs: For now lets leave it spring.cloud.task.events...

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public ListenerProperties getSkip() {
+        return skip;
+    }
+
+    public void setSkip(ListenerProperties skip) {
+        this.skip = skip;
+    }
+
+    public  class ListenerProperties {
```
cppwfs: Why do we need this ListenerProperties inner class?

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.core.Ordered;
+
+/**
+ *  @author Ali Shahbour
+ */
+@ConfigurationProperties(prefix = "spring.cloud.task.events")
+public class TaskEventProperties {
+
+    private ListenerProperties jobExecution = new ListenerProperties(Ordered.HIGHEST_PRECEDENCE);
```
cppwfs: Add comments to each property.  i.e. https://github.com/spring-cloud/spring-cloud-task/blob/master/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java#L32

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
```diff
@@ -0,0 +1,93 @@
+package org.springframework.cloud.task.batch.listener.support;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.core.Ordered;
+
+/**
+ *  @author Ali Shahbour
+ */
+@ConfigurationProperties(prefix = "spring.cloud.task.events")
```
cppwfs: you were correct in that prefix should be spring.cloud.task.batch.events.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingStepExecutionListener.java
```diff
@@ -30,16 +31,24 @@
...
 
 	private MessagePublisher<StepExecutionEvent> messagePublisher;
+	private int order = Ordered.HIGHEST_PRECEDENCE;
 
 	public EventEmittingStepExecutionListener(MessageChannel output) {
 		Assert.notNull(output, "An output channel is required");
 		this.messagePublisher = new MessagePublisher<>(output);
 	}
 
+	public EventEmittingStepExecutionListener(MessageChannel output,int order) {
```
cppwfs: formatting:  space after comma.  i.e. `, int order`

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingChunkListener.java
```diff
@@ -0,0 +1,60 @@
+package org.springframework.cloud.task.batch.listener;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.batch.core.ChunkListener;
+import org.springframework.batch.core.ItemReadListener;
```
cppwfs: Cleanup imports.  Unused imports are present.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingChunkListener.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ *  The {@link ChunkListener#beforeChunk()} and
+ *  {@link ChunkListener#afterChunk(Object)} are both no-ops in this implementation.
+ *  {@link ChunkListener#afterChunkError(Object)}.
+ *
+ * @author Ali Shahbour
+ */
+public class EventEmittingChunkListener implements ChunkListener, Ordered {
+
+    private static final Log logger = LogFactory.getLog(EventEmittingChunkListener.class);
```
cppwfs: Use tabs not spaces.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
```diff
@@ -0,0 +1,93 @@
+package org.springframework.cloud.task.batch.listener.support;
```
cppwfs: Used tabs instead of spaces.

cppwfs:
Also need to add usage documentation to the https://github.com/spring-cloud/spring-cloud-task/blob/master/spring-cloud-task-docs/src/main/asciidoc/stream.adoc.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
shahbour: the question is why is it inner class or why we added this class in the first place ?

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/BatchEventAutoConfiguration.java
```diff
@@ -77,59 +81,58 @@ public TaskBatchEventListenerBeanPostProcessor batchTaskEventListenerBeanPostPro
...
 	@EnableBinding(BatchEventsChannels.class)
+	@EnableConfigurationProperties(TaskEventProperties.class)
 	@ConditionalOnMissingBean(name = JOB_EXECUTION_EVENTS_LISTENER)
 	public static class JobExecutionListenerConfiguration {
 
 		@Autowired
 		private BatchEventsChannels listenerChannels;
 
+		@Autowired
+		TaskEventProperties taskEventProperties;
```
mminella: Why is this package private?

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
```diff
@@ -0,0 +1,93 @@
...
+@ConfigurationProperties(prefix = "spring.cloud.task.events")
+public class TaskEventProperties {
+
+    private ListenerProperties jobExecution = new ListenerProperties(Ordered.HIGHEST_PRECEDENCE);
+    private ListenerProperties stepExecution = new ListenerProperties(Ordered.HIGHEST_PRECEDENCE);
+    private ListenerProperties itemRead = new ListenerProperties(Ordered.HIGHEST_PRECEDENCE);
+    private ListenerProperties itemProcess = new ListenerProperties(Ordered.HIGHEST_PRECEDENCE);
+    private ListenerProperties itemWrite = new ListenerProperties(Ordered.HIGHEST_PRECEDENCE);
+    private ListenerProperties chunk = new ListenerProperties(Ordered.HIGHEST_PRECEDENCE);
+    private ListenerProperties skip = new ListenerProperties(Ordered.HIGHEST_PRECEDENCE);
```
mminella: Why highest and not lowest?

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
mminella: Javadoc across the class.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
mminella: +1 on the need for this class in general.

cppwfs: In short I'm not sure we need this inner class since it only stores the order.

shahbour: i could not figure out how to read all the properties 
spring.cloud.task.batch.events:

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingChunkListener.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ *  Provides informational messages around the {@link Chunk} of a batch job.
+ *
+ *  The {@link ChunkListener#beforeChunk()} and
+ *  {@link ChunkListener#afterChunk(Object)} are both no-ops in this implementation.
+ *  {@link ChunkListener#afterChunkError(Object)}.
+ *
+ * @author Ali Shahbour
+ */
+public class EventEmittingChunkListener implements ChunkListener, Ordered {
```
mminella: What is the purpose of this class?

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
shahbour: i think highest was requested by @cppwfs  because batch run the listeners in reverse order

cppwfs:
no.  I suggested the lowest.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
mminella: That depends.  I believe (I'd have to double check) that we run in order on the way in and reverse on the way out (I actually think there is a batch PR to address the consistency of this rule).

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/EventEmittingChunkListener.java
shahbour: To implement Ordered i had to create a `EventEmittingChunkListener` , currently it does not emit any thing but we can just send event as reader .

mminella: My point is that I believe the original option for the chunk listener did emit messages (via the proxy) so I'd expect if we're going down this route, we'd implement that functionality.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/support/TaskEventProperties.java
shahbour: In my project i did override the AutoConfiguration and used Highest to run this listener at the end after my own project listener

shahbour:
@cppwfs  you asked for Integer.MIN_VALUE that is the same as HIGHEST_PRECEDENCE

shahbour:
i pushed some changes requested tomorrow i will check changes required by @mminella

cppwfs:
@shahbour My bad on that.

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/BatchEventAutoConfiguration.java
shahbour: i did not get you , i should change TaskEventProperties to private ?

shahbour:
For both classes `TaskEventProperties` and `ListenerProperties` the reason i did create two classes was because of requirement 
```
spring.cloud.task.batch.events.job-execution.order=5
spring.cloud.task.batch.events.step-execution.order=5
spring.cloud.task.batch.events.chunk.order=5
spring.cloud.task.batch.events.item-read.order=5
spring.cloud.task.batch.events.item-process.order=5
spring.cloud.task.batch.events.item-write.order=5
spring.cloud.task.batch.events.skip.order=5
```  
so the only way i found it possible to read all this different values is to create two classes which seems too much . 

Can you please guide me on what you think is better to read these values ? Should i use Map<Map<>> ?

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/batch/listener/BatchEventAutoConfiguration.java
mminella: Yes unless there is a reason it should be exposed.

shahbour:
I tried removing the extra class `ListenerProperties` and using 
```
/**
* HashMap to hold all properties for jobExecution
*/
private Map<String, Integer> jobExecution = new HashMap<>();
```
but that did not work because we already have enabled property with boolean 
```
spring.cloud.task.batch.events.job-execution.enabled=true
spring.cloud.task.batch.events.job-execution.order=5
```

So the only solution i see is to keep the `ListenerProperties` or change the map into Map<String,Object> and then do casting ?

What do you think , also for the priority should i keep it as highest or change it to lowest ?

cppwfs:
To the first question:
You can simplify it to something like: spring.cloud.task.batch.events.jobExecutionOrder, spring.cloud.task.batch.events.stepExecutionOrder.  And this should negate the need for the inner class etc.

To the second question: 
It should be LOWEST_PRECEDENCE.  Sorry for the confusion.

shahbour:
any other requirements for this PR

spring-cloud-task-stream/src/test/java/org/springframework/cloud/task/batch/listener/EventListenerTests.java
```diff
@@ -58,9 +62,22 @@ public void beforeTests() {
...
 
 	@Test
+	public void testEventListenerOrderProperty() {
+		assertEquals(eventEmittingSkipListener.getOrder(),Ordered.LOWEST_PRECEDENCE);
+		assertEquals(eventEmittingItemProcessListener.getOrder(), Ordered.LOWEST_PRECEDENCE);
+		assertEquals(eventEmittingItemReadListener.getOrder(), Ordered.LOWEST_PRECEDENCE);
+		assertEquals(eventEmittingItemWriteListener.getOrder(), Ordered.LOWEST_PRECEDENCE);
+		assertEquals(eventEmittingJobExecutionListener.getOrder(), Ordered.LOWEST_PRECEDENCE);
+		assertEquals(eventEmittingStepExecutionListener.getOrder(), Ordered.LOWEST_PRECEDENCE);
+		assertEquals(eventEmittingChunkListener.getOrder(),0);
```
cppwfs: Was wondering why we are setting chunk listener to 0 instead of LOWEST_PRECEDENCE?  I can change it to LOWEST_PRECEDENCE upon merge if you agree.

shahbour: The idea is to confirm two parameters constructor is setting the order and not getting the default value. Feel free to change it if you want

cppwfs:
LGTM.  Waiting to merge this till after the M2 release which should be early next week.

cppwfs:
A little cleanup.  Rebased, Squashed Merged.
Thank you for your contribution!

PR #296, opened 2017-03-09
Filter out executions with null in TASK_NAME

### LINKED ISSUE #295: Filter out executions with null in TASK_NAME
Looks like launching older tasks that aren't aware that they should update the task executions rather than inserting a new row results in orphaned task execution rows with null in all fields except for ID and EXTERNAL_EXECUTION_ID. We should filter these out when listing all task executions.

See: https://github.com/spring-cloud/spring-cloud-dataflow/issues/1201

trisberg:
Resolves #295


mminella:
There's an ongoing conversation on this issue in the issue itself.  Let's address that before merging this PR.

trisberg:
Closing this since it's no longer relevant once we add task name: https://github.com/spring-cloud/spring-cloud-task/pull/299

PR #300, opened 2017-03-29
Refactor to clense task name

### LINKED ISSUE #280: Context IDs may be colon delimited which breaks JMX.
In `SimpleTaskNameResolver`, if a name isn't configured, we use the current application context's id as the name.  If a profile is in use, the resulting id may have a colon (`:`) in it.  This is not allowed and causes an exception when using JMX.  We should auto-convert the colons into something allowed (forward slash, etc).

mminella:
When using Spring Boot, the id of the application context can be a colon
delimited string of values.  By default, Spring Cloud Task uses this
value as the task name.  However, when using JMX, this can cause issues.
This commit updates the SimpleTaskNameResolver to clense the colons and
replace them with underscores.

Resolves #280


cppwfs:
LGTM

cppwfs:
Rebased Merged

PR #301, opened 2017-03-30
Removed dependencies on Spring Boot starters

### LINKED ISSUE #294: Please don't depend on Starter POMs
I think libraries like `spring-cloud-starter-core` and `spring-cloud-task-stream` should not depend on Spring Boot Starter POMs.

For various reasons some of my projects can not make the switch to logback as a logging backend, so I am forced to exclude `spring-boot-starter-logging` whenever I want to use `spring-cloud-task-core`.

Please only depend on those libraries and APIs you really use. If you want to provide a quick start for users of spring-cloud-task, then create additional `spring-cloud-task-*-starter`s.

mminella:
This commit removes the dependency on any Spring Boot starters.
Starters are still used in the test scope, but should not impact the
library's dependencies.

Resolves #294

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,6 +117,11 @@
 				<artifactId>spring-cloud-stream-test-support</artifactId>
 				<version>${spring-cloud-stream.version}</version>
 			</dependency>
+			<dependency>
+				<groupId>commons-logging</groupId>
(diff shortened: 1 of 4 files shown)


trisberg:
LGTM, merging in a few ...

PR #302, opened 2017-04-10
Change closecontextEnable to closeContextEnabled

### LINKED ISSUE #217: closecontext.enable property needs to be set to closecontext.enabled
To be consistent with boot spring.cloud.task.closecontext.enable needs to be renamed to spring.cloud.task.closecontext.enabled .  And we need to follow the deprecation process for this.

dturanski:
This fixes #217 . The old property name still works but is deprecated. Will log a warning if used. The docs updated to reference the new property.

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
@@ -16,17 +16,21 @@
 
 package org.springframework.cloud.task.configuration;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 /**
  * Properties available to configure the task.
  *
  * @author Glenn Renfro
+ * @author David Turanski
  */
 
 @ConfigurationProperties(prefix = "spring.cloud.task")
 public class TaskProperties {
-
+	private static final Logger logger = LoggerFactory.getLogger(TaskProperties.class);
+	
 	public static final String DEFAULT_TABLE_PREFIX = "TASK_";
 
 	/**
@@ -54,7 +58,7 @@ public class TaskProperties {
 	 * When set to true the context is closed at the end of the task.  Else
 	 * the context remains open.
 	 */
-	private Boolean closecontextEnable = true;
+	private Boolean closecontextEnabled = true;
 
 	public String getExternalExecutionId() {
 		return externalExecutionId;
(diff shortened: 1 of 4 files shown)


spring-cloud-task-core/src/test/java/org/springframework/cloud/task/configuration/TaskPropertiesTests.java
```diff
@@ -0,0 +1,55 @@
...
+		TaskExecutionDaoFactoryBean taskExecutionDaoFactoryBean() {
+			return new TaskExecutionDaoFactoryBean();
+		}
+		@Bean
+		public TaskRepository taskRepository(TaskExecutionDaoFactoryBean tefb) {
+			return new SimpleTaskRepository(tefb);
+		}
+	}
+}
+
```
cppwfs: Add blank line at end of class.  Will add on merge

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/configuration/TaskPropertiesTests.java
```diff
@@ -0,0 +1,55 @@
...
+@SuiteClasses({
+		TaskPropertiesTests.CloseContextEnableTest.class,
+		TaskPropertiesTests.CloseContextEnabledTest.class,
+		
+})
+
+@DirtiesContext
+public class TaskPropertiesTests {	
+	@Test
+	public void test() {
```
cppwfs: A tad picky but we can also check the property itself.   For example:
```
	@Autowired
	TaskProperties taskProperties;
	
	@Test
	public void test() {
		assertThat(taskProperties.getClosecontextEnabled(), is(false));
	}
```

I can add during merge if you have no objections.

spring-cloud-task-docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -145,7 +145,7 @@ Spring Cloud Task itself:
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-task-core</artifactId>
-			<version>1.0.0.RELEASE</version>
+			<version>1.2.0.BUILD-SNAPSHOT</version>
```
cppwfs: thanks for the doc update!

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
```diff
@@ -16,17 +16,21 @@
 
 package org.springframework.cloud.task.configuration;
 
+import org.slf4j.Logger;
```
cppwfs: Spring Cloud Task and SCDF now use apache commons.
```
import org.apache.commons.logging.Log;
import org.apache.commons.logging.LogFactory;
...
private final static Log logger = LogFactory.getLog(TaskProperties.class);
...
```
Will fix on merge.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/configuration/TaskPropertiesTests.java
dturanski: good idea

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
dturanski: wow! Why is that?

cppwfs: There were conversations on it.  But I was not a part of them. :-(

PR #310, opened 2017-05-04
Task identifies if JPA is in use and uses JPATransactionManager

### LINKED ISSUE #309: Support JPATransactionManager in the DefaultConfigurer
DefaultTaskConfigurer should auto-detect that JPA is in use and set the TransactionManager to use JPATransactionManager.

cppwfs:
resolves #309


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
```diff
@@ -157,11 +157,11 @@ private TaskConfigurer getDefaultConfigurer() {
...
 				taskConfigurer = new DefaultTaskConfigurer(
 						this.dataSources.iterator().next(),
-						taskProperties.getTablePrefix());
+						taskProperties.getTablePrefix(), context);
 			}
 			else {
-				taskConfigurer = new DefaultTaskConfigurer(
-						taskProperties.getTablePrefix());
+				taskConfigurer = new DefaultTaskConfigurer( null,
+						taskProperties.getTablePrefix(), null);
```
mminella: The old constructor is still there, why did this need to be changed?

spring-cloud-task-samples/jpa-sample/README.adoc
```diff
@@ -0,0 +1,28 @@
+= Spring Batch Job Task
```
mminella: Fix title

spring-cloud-task-samples/jpa-sample/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+			<artifactId>hibernate-entitymanager</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>com.h2database</groupId>
+			<artifactId>h2</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.mariadb.jdbc</groupId>
+			<artifactId>mariadb-java-client</artifactId>
+		</dependency>
```
mminella: Why are we requiring both maria and h2 in the sample?

spring-cloud-task-samples/jpa-sample/src/main/java/io/spring/JPAApplication.java
```diff
@@ -0,0 +1,31 @@
...
+
+package io.spring;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.cloud.task.configuration.EnableTask;
+
+@SpringBootApplication
+@EnableTask
+public class JPAApplication {
```
mminella: This should be camel cased `JpaApplication`.  See `JpaTransactionManager` as an example of the precedent: http://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/orm/jpa/JpaTransactionManager.html  The README will also need to be updated.

spring-cloud-task-samples/jpa-sample/src/test/java/io/spring/JpaApplicationTests.java
```diff
@@ -0,0 +1,108 @@
...
+		dataSource.setPassword(DATASOURCE_USER_PASSWORD);
+		this.dataSource = dataSource;
+		try {
+			this.server = Server.createTcpServer("-tcp", "-tcpAllowOthers", "-tcpPort", String.valueOf(randomPort))
+					.start();
+		}
+		catch (SQLException e) {
+			throw new IllegalStateException(e);
+		}
+	}
```
mminella: Why are we creating a db this way instead of just using the embedded option?

cppwfs: The application is already finished by the time the tests can be run.   So we need an externally run database that is still active so that we can test the results

spring-cloud-task-samples/jpa-sample/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+			<artifactId>spring-cloud-starter-task</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-data-jpa</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-cloud-connectors</artifactId>
+		</dependency>
```
cppwfs: Removed

mminella:
Merged as b03f8398a7a367153de6c2a9fa86c4e58e2f85fb

PR #312, opened 2017-05-04
Repository is not initialized if tablePrefix is set.

### LINKED ISSUE #311: If table prefixes are set do not initialize tables in datastore
If a user sets the table prefixes do not initialize the tables in the datastore.

cppwfs:
resolves #311

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
@@ -25,6 +25,7 @@ import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.ApplicationArguments;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.task.listener.TaskLifecycleListener;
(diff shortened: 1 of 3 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskConfiguration.java
```diff
@@ -111,6 +112,7 @@ public TaskNameResolver taskNameResolver() {
 	}
 
 	@Bean
+	@ConditionalOnProperty(prefix = "spring.cloud.task", name = "tablePrefix", matchIfMissing = true)
```
mminella: How does this prevent this from being initialized?  This looks like if the `spring.cloud.task.tablePrefix` is there, it will create the bean.  Can we get a test to prove this works?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskRepositoryInitializer.java
```diff
@@ -86,7 +89,7 @@ private String getDatabaseType(DataSource dataSource) {
 
 	@Override
 	public void afterPropertiesSet() throws Exception {
-		if (dataSource != null && taskInitializationEnable) {
+		if (dataSource != null && taskInitializationEnable && this.tablePrefix == null) {
```
mminella: Probably should be `StringUtils.hasText(this.tablePrefix)`.  Will fix on merge.

mminella:
Merged as 6c47223859e1d0d709557f9c74c52f8f2f743480

PR #317, opened 2017-06-08
Sct 313

cppwfs:


cppwfs:
Closing to sync up versions with waffle.

PR #335, opened 2017-09-19
Prevent a task from starting if an instance with the same name is run…

### LINKED ISSUE #81: Restricting Spring Cloud Task instances
There are use cases where an application is triggered by cron scheduler but due to communication with 3rd party API(s), the application runs slow and hence the cron scheduler triggers another instance of the application again while the first instance is still running.

cppwfs:
…ning.

feature is disabled by default.  To enable it set spring.cloud.task.singleInstanceEnabled=true.
resolves #81

added docs

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
@@ -62,6 +62,16 @@ public class TaskProperties {
 	 */
 	private Boolean closecontextEnabled = true;
 
+	/**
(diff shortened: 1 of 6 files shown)


mminella:
1. Why doesn't this use Spring Integration's distributed locking mechanism?
2. Why is this not just a regular `TaskExecutionListener`?

cppwfs:
1) Can you provide me a link to what you referring?  Is it the JdbcLockRegistry?
2) it is a core feature of the task lifecycle.  i.e. don't start if the there is another task execution with the same name is running.

cppwfs:
Closing this PR and will create a new one.

PR #344, opened 2017-10-06
SCT now handles exceptions thrown by execution listeners

### LINKED ISSUE #341: Task should be able to record end of task info if exception is fired during TaskExecutionListener.

cppwfs:
resolves #341

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
@@ -256,12 +256,12 @@ public class TaskLifecycleListener implements ApplicationListener<ApplicationEve
 					taskExecutionListener.onTaskStartup(listenerTaskExecution);
 				}
 			}
-			catch (Throwable listenerException) {
-				logger.warn(listenerException);
+			catch (Throwable currentListenerException) {
+				logger.warn(currentListenerException);
 				this.listenerFailed = true;
-				this.taskExecution.setErrorMessage(listenerException.getMessage());
-				this.listenerException = listenerException;
-				throw listenerException;
+				this.taskExecution.setErrorMessage(currentListenerException.getMessage());
+				this.listenerException = currentListenerException;
+				throw currentListenerException;
 			}
 		}
 		return listenerTaskExecution;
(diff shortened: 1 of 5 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -165,18 +169,34 @@ else if(this.applicationFailedEvent != null){
...
+				setMessages(invokeOnTaskError(this.taskExecution,
+						this.applicationFailedEvent.getException()));
+			}
 
-			if (this.applicationFailedEvent != null && this.taskExecution.getExitCode() != 0) {
-				taskExecution.setExitMessage(invokeOnTaskError(taskExecution,
-						this.applicationFailedEvent.getException()).getExitMessage());
+			setMessages(invokeOnTaskEnd(this.taskExecution));
+			if(this.listenerFailed) {
+				this.taskExecution.setExitCode(1);
```
mminella: Should this be hardcoded as 1, a specific value that we specify other than one (to indicate the specific error condition being an exception was thrown via a SCT listener), or should this be the result of any configured `ExitCodeGenerator`?  Where in the lifecycle does the `ExitCodeGenerator` get called based on this logic?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -220,24 +245,44 @@ private void doTaskStart() {
...
+				for (TaskExecutionListener taskExecutionListener : this.taskExecutionListeners) {
+					taskExecutionListener.onTaskStartup(listenerTaskExecution);
+				}
+			}
+			catch (Throwable listenerException) {
+				logger.warn(listenerException);
+				this.listenerFailed = true;
+				this.taskExecution.setErrorMessage(listenerException.getMessage());
+				this.listenerException = listenerException;
+				throw listenerException;
```
mminella: Bad practice to use duplicate variable names.  Please use something different.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
cppwfs: With this PR we stop processing all TaskExecutionListeners when an exception is thrown by a listener.  So if I understand you correctly, you want to check to see if the exception thrown by the listener implements the ExitCodeGenerator interface and if so use that exit code.

mminella: No.  I wouldn't expect an exception to implement `ExitCodeGenerator`.  However, there may be one in the `ApplicationContext`.  I'm asking if we should use it...

cppwfs:
Documentation updated as well.

mminella:
Rebased, squashed, and merged as 3f316a01708cb59f8c73c365b5a242217c25cb13

PR #345, opened 2017-10-12
Prevent a task from starting if an inst with the same name is running.

### LINKED ISSUE #81: Restricting Spring Cloud Task instances

cppwfs:
resolves #81

Using LockRegistryLeaderInitiator to do leadership election.

When task is started and singleInstanceEnabled isset to true then we use leader election
to determine if a task needs to be started.

May want to wait for the #341 to be merged first.


cppwfs:
Because of a dependency on Spring Integration, do not merge until SI M8 has been released.   The PR that this is dependent on is: https://github.com/spring-projects/spring-integration/pull/2265.
It has been merged and is available via snapshot.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SingleInstanceTaskListener.java
```diff
@@ -0,0 +1,151 @@
...
+		this.lockRegistryLeaderInitiator = new LockRegistryLeaderInitiator(
+				this.lockRegistry,
+				new DefaultCandidate(String.valueOf(taskExecution.getExecutionId()),
+						taskNameResolver.getTaskName()));
+		this.lockRegistryLeaderInitiator.setApplicationEventPublisher(this.applicationEventPublisher);
+		this.lockRegistryLeaderInitiator.setPublishFailedEvents(true);
+		this.lockRegistryLeaderInitiator.start();
+		while (!this.lockReady) {
+			try {
+				Thread.sleep(500);
```
mminella: Is there a reason this isn't configurable?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
```diff
@@ -62,6 +62,23 @@
 	 */
 	private Boolean closecontextEnabled = true;
 
+	/**
+	 * When set to true the
+	 * {@link org.springframework.cloud.task.listener.TaskLifecycleListener#doTaskStart()}
```
mminella: We shouldn't be quite so specific on the implementation details in javadoc.  Say we wanted to change the implementation...how we cause this to happen is independent of the fact that it does.  Since javadoc is part of the contract, we need to be careful with how specific we are.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
```diff
@@ -62,6 +62,23 @@
...
+	 * will check to see if a task execution with the same task name is already
+	 * running.  If a task is still running then it will throw a
+	 * {@link org.springframework.cloud.task.listener.TaskExecutionException}.
+	 * When task execution ends the lock is released.
+	 */
+	private Boolean singleInstanceEnabled = false;
```
mminella: Why `Boolean` and not `boolean`?  Would this ever be null (and be ok)?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
```diff
@@ -123,4 +140,20 @@ public Long getParentExecutionId() {
 	public void setParentExecutionId(Long parentExecutionId) {
 		this.parentExecutionId = parentExecutionId;
 	}
+
+	public Boolean getSingleInstanceEnabled() {
+		return singleInstanceEnabled;
+	}
```
mminella: What is the purpose of the getter?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
```diff
@@ -123,4 +140,20 @@ public Long getParentExecutionId() {
...
+
+	public void setSingleInstanceEnabled(Boolean singleInstanceEnabled) {
+		this.singleInstanceEnabled = singleInstanceEnabled;
+	}
```
mminella: See previous comment about `Boolean` vs `boolean`.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
```diff
@@ -123,4 +140,20 @@ public Long getParentExecutionId() {
...
+
+	public int getSingleInstanceLockTtl() {
+		return singleInstanceLockTtl;
+	}
```
mminella: Again why the getter?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SingleInstanceTaskListener.java
```diff
@@ -0,0 +1,151 @@
...
+import org.springframework.integration.leader.DefaultCandidate;
+import org.springframework.integration.leader.event.OnFailedToAcquireMutexEvent;
+import org.springframework.integration.leader.event.OnGrantedEvent;
+import org.springframework.integration.support.leader.LockRegistryLeaderInitiator;
+import org.springframework.integration.support.locks.LockRegistry;
+
+/**
+ * When singleInstanceEnabled is set to true this listener will create a lock for the task
+ * based on the spring.cloud.task.name. If a lock already exists this Listener will throw
+ * a TaskExecutionException.
```
mminella: We should document that if this is added manually, it should be the first listener in the chain.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
cppwfs: Consistency

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java
cppwfs: consistency

spring-cloud-task-core/pom.xml
```diff
@@ -60,6 +60,16 @@
...
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-core</artifactId>
+			<version>5.0.0.RELEASE</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-jdbc</artifactId>
+			<version>5.0.0.RELEASE</version>
+		</dependency>
```
mminella: Shouldn't these be optional dependencies?  We don't want to pull in SI if they aren't using this feature...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleSingleTaskConfiguration.java
```diff
@@ -0,0 +1,78 @@
...
+ * Autoconfiguration of {@link SingleInstanceTaskListener}.
+ *
+ * @author Glenn Renfro
+ * @since 2.0.0
+ */
+
+@Order(Ordered.HIGHEST_PRECEDENCE)
+@Configuration
+@ConditionalOnProperty(prefix = "spring.cloud.task", name = "singleInstanceEnabled", havingValue = "true")
+public class SimpleSingleTaskConfiguration {
```
mminella: I'd drop the `Simple` from this class name...

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleSingleTaskConfiguration.java
```diff
@@ -0,0 +1,78 @@
...
+
+	@Autowired(required = false)
+	Collection<DataSource> dataSources;
```
mminella: This is never used.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleSingleTaskConfiguration.java
```diff
@@ -0,0 +1,78 @@
...
+
+	@Autowired
+	private TaskProperties taskProperties;
+
+	@Autowired
+	private SimpleTaskNameResolver taskNameResolver;
+
+	@Autowired
+	private ConfigurableApplicationContext context;
```
mminella: This is never used.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SingleInstanceTaskListener.java
```diff
@@ -0,0 +1,154 @@
...
+	public void onApplicationEvent(ApplicationEvent applicationEvent) {
+		if (applicationEvent instanceof OnGrantedEvent) {
+			this.lockReady = true;
+		}
+		else if (applicationEvent instanceof OnFailedToAcquireMutexEvent) {
+			this.lockFailed = true;
+		}
+	}
+
+	private LockRegistry getDefaultLockRegistry( long executionId) {
```
mminella: Is there a reason for this being lazily created and not just done in the constructor that doesn't set a registry?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SingleInstanceTaskListener.java
```diff
@@ -0,0 +1,154 @@
...
+import org.springframework.integration.jdbc.lock.DefaultLockRepository;
+import org.springframework.integration.jdbc.lock.JdbcLockRegistry;
+import org.springframework.integration.leader.DefaultCandidate;
+import org.springframework.integration.leader.event.OnFailedToAcquireMutexEvent;
+import org.springframework.integration.leader.event.OnGrantedEvent;
+import org.springframework.integration.support.leader.LockRegistryLeaderInitiator;
+import org.springframework.integration.support.locks.LockRegistry;
+
+/**
+ * When singleInstanceEnabled is set to true this listener will create a lock for the task
```
mminella: "When `spring.cloud.task.singleInstanceEnabled` is set to `true`"...please use the full property name as well as formatting

spring-cloud-task-integration-tests/src/test/java/org/springframework/cloud/task/executionid/TaskStartTests.java
```diff
@@ -207,6 +247,16 @@ private boolean waitForDBToBePopulated() throws Exception {
 		return isDbPopulated;
 	}
 
+	private void enableFooLock() {
+		SimpleJdbcInsert taskLockInsert = new SimpleJdbcInsert(this.dataSource).withTableName("TASK_LOCK");
+		Map<String, Object> taskLockParams = new HashMap<>();
+		taskLockParams.put("LOCK_KEY", "acbd18db-4cc2-385c-adef-654fccc4a4d8");
+		taskLockParams.put("REGION", "DEFAULT");
+		taskLockParams.put("CLIENT_ID", "ac39193d-4a24-4112-9d1b-5d79d6286921");
```
mminella: Is there a reason these UUIDs are hardcoded?

spring-cloud-task-integration-tests/src/test/java/org/springframework/cloud/task/executionid/TaskStartTests.java
```diff
@@ -207,6 +247,16 @@ private boolean waitForDBToBePopulated() throws Exception {
 		return isDbPopulated;
 	}
 
+	private void enableFooLock() {
+		SimpleJdbcInsert taskLockInsert = new SimpleJdbcInsert(this.dataSource).withTableName("TASK_LOCK");
```
mminella: I had never seen `SimpleJdbcInsert` before.  Cool feature!

spring-cloud-task-integration-tests/src/test/java/org/springframework/cloud/task/executionid/TaskStartTests.java
cppwfs: The lock key is generated on the task name `foo`.  And that is how it keyed.  as far as the client id, it was just an id that was used.

spring-cloud-task-core/pom.xml
cppwfs: If we set them as optional we get class not found exception. Are we saying that the user has to explicitly add the SI dependency if they want to use this feature?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SingleInstanceTaskListener.java
cppwfs: We use the executionId as a unique id for the lock registry.  This is not available at constructor time.

spring-cloud-task-core/pom.xml
mminella: Well we can’t just bring in SI for applications that aren’t going to use it.  I’d expect a user that was going to use this functionality to use the `spring-cloud-starter-task` and the `spring-boot-starter-integration`.  We’d need to document that it’s needed, but if an app isn’t going to use SI, we can’t just require the dependency

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SingleInstanceTaskListener.java
mminella: I missed that. Thanks.

spring-cloud-task-integration-tests/src/test/java/org/springframework/cloud/task/executionid/TaskStartTests.java
mminella: I’d just expect to use `UUID.randomUUID()` to prevent the potential for duplicates across test runs.

cppwfs: The lock key is created using SI's UUIDConverter, which will generate a UUID from a key passed in.  I will update the test so that it will accept a key and create its UUID from it.

mminella:
Merged as d8a73ba1835e83de8568dc0c3569916858d85987.  This cannot be backported to 1.2.x due to the dependency on Spring Integration 5.

PR #368, opened 2018-01-05
Updated tests to remove deprecated code from tests

### LINKED ISSUE #338: Need to remove deprecations from tests
new PageRequest needs to be replaced with PageRequest.of

cppwfs:
* cleanup removing unused headers
* Updated asserts in code base that needed messages (marked as deprecated)
* left one test that was testing a deprecated constructor.  When that code is removed we can remove that test.
* some other cleanup

resolves #338

### DIFF
--- a/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/SimpleCommandLineArgsProviderTests.java
+++ b/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/SimpleCommandLineArgsProviderTests.java
@@ -30,6 +30,7 @@ import static org.junit.Assert.assertEquals;
  */
 public class SimpleCommandLineArgsProviderTests {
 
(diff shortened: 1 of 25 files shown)


spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/SimpleCommandLineArgsProviderTests.java
```diff
@@ -30,6 +30,7 @@
  */
 public class SimpleCommandLineArgsProviderTests {
 
+	//TODO remove when SimpleCommandLineArgsProvider is removed.
```
mminella: Why would `SimpleCommandLineArgsProvider` be removed?  Do you mean the constructor?

mminella:
Rebased and merged as 3ad9efe3febdef0d67a7e5fcbcc1ec5dca5f0969.

PR #371, opened 2018-01-05
Removed code that was marked as deprecated.

### LINKED ISSUE #366: Remove Deprecations from the 1.x line
TaskExecutionDaoFactoryBean.setTablePrefix
JdbcTaskBatchDao.setTablePrefix
SimpleCommndLIneArgsProvider.Constructor
JdbcTaskExecutionDao.setTAblePrefix
etc.

cppwfs:
Also updated test that tested deprecated code.

resolves #366

### DIFF
--- a/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
+++ b/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
@@ -222,8 +222,7 @@ public class DeployerPartitionHandler implements PartitionHandler, EnvironmentAw
 
 		if(this.commandLineArgsProvider == null) {
 			SimpleCommandLineArgsProvider provider = new
-					SimpleCommandLineArgsProvider();
-			provider.onTaskStartup(taskExecution);
+					SimpleCommandLineArgsProvider(taskExecution);
 			this.commandLineArgsProvider = provider;
 
 		}
--- a/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
+++ b/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
@@ -221,8 +221,11 @@ public class DeployerPartitionHandler implements PartitionHandler, EnvironmentAw
 		this.taskExecution = taskExecution;
 
 		if(this.commandLineArgsProvider == null) {
-			this.commandLineArgsProvider =
-					new SimpleCommandLineArgsProvider(this.taskExecution);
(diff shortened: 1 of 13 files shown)


spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
```diff
@@ -221,8 +221,11 @@ public void beforeTask(TaskExecution taskExecution) {
 		this.taskExecution = taskExecution;
 
 		if(this.commandLineArgsProvider == null) {
-			this.commandLineArgsProvider =
-					new SimpleCommandLineArgsProvider(this.taskExecution);
+			SimpleCommandLineArgsProvider provider = new
+					SimpleCommandLineArgsProvider();
+			provider.onTaskStartup(taskExecution);
```
mminella: Why is this being created this way instead of using the listener correctly?  This implies that the deprication doesn't make sense in the first place...

cppwfs: So should we just remove the deprecation?

mminella: Why was it depricated?  We need to do the homework as to why it was.  I'd assume that we moved in this direction so that Spring would take care of the injection of the task execution.  If that's the case, shouldn't the provider be registered as a listener and injected here instead of newing up a new one?

cppwfs: IMHO I think we should remove the deprecation itself and allow the Constructor to set the TaskExecution.  The SimpleCommandLineArgsProvider extends TaskExecutionListenerSupport that means that `onTaskStartup` its TaskExecution would get set.   However, SimpleCommandLineArgsProvider is never created as a bean and thus the `onTaskStartup` is never called.   And SimpleCommandLineArgsProvider is only used by the DeployerPartitionHandler, if a CommandLineArgsProvider has not been provided to the DeployerPartitionHandler.

mminella: I don't have an argument against that so make it so :)

mminella:
I remember now (as I read it) why we removed the constructor and added the listener support.  If I want to configure the `SimpleCommandLineArgsProvider` myself, I won't have a `TaskExecution` when the context is created.  It must be populated by the `TaskExecutionListener`.  I think the depreciation was too aggressive and both the constructor option and the listener option make sense.

mminella:
Rebased, squashed, polished, and merged as 76309b89ffc8ad7bb7f0035c235a23c73a8672ad.

PR #375, opened 2018-01-12
Adds configuration metadata for properties

### LINKED ISSUE #365: please add the `spring-boot-configuration-processor` to support generate configuration metadata
If you add it to your build, as [per the docs](https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html#configuration-metadata-annotation-processor). it'll generate a `.json` file that all of the big tools (STS, NetBeans, IntelliJ and even VSCode thanks to our language servers in STS4) know how to use to support auto completion in `application.properties`. it would be nice if properties like `spring.cloud.task.closecontext_enabled` were discoverable through this metadata.

cppwfs:
resolves #365

### DIFF
--- a/spring-cloud-task-batch/pom.xml
+++ b/spring-cloud-task-batch/pom.xml
@@ -65,5 +65,10 @@
 			<artifactId>h2</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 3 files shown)


mminella:
We need this in the other modules that have properties (batch and stream both have properties as well...).

cppwfs:
Good Catch.  Updated

mminella:
LGTM. Merged as fd2d4a91245a5dbfd7743a04869ed41129d3a715.

PR #377, opened 2018-01-18
Set the Spring-Cloud-Build to Snapshot until we go RC1.

### LINKED ISSUE #360: Moving the `ApplicationReadyEvent` to before the execution of *Runners breaks Spring Cloud Task
Prior to issue https://github.com/spring-projects/spring-boot/issues/7656, Spring Cloud Task was able to use the `ApplicationReadyEvent` to be notified that the `ApplicationRunner`s and `CommandLineRunner`s have been executed.  The resolution of that issue breaks Spring Cloud Task in that it moves that event to before the runners have been executed without providing a replacement notification.

cppwfs:
Did build and smoke tests and it resolves the issues identified in this issue.
resolves #360

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
-		<version>2.0.0.M4</version>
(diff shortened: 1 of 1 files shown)


mminella:
LGTM.  Rebased and merged as 6860656d98e9f968f5b3011699a3347c132c8bf9.

PR #381, opened 2018-01-23
Set the spring.cloud.task.closecontext_enable default to false

### LINKED ISSUE #364: Re-evaluate the default of the `spring.cloud.task.closecontext_enabled` flag
This flag was originally defaulted to `true` to allow users to use Spring Boot's Acuator within a Spring Cloud Task (programmatically shutting down the context) when Acuator required Tomcat.  This is no longer the case so we can change this default to `false`, allowing projects generated at start.spring.io to build successfully without modification.

cppwfs:
resolves #364


mminella:
LGTM.  Rebased and merged as 94e074d8412b9cf3163c125292de2d3d5858f2c7.

PR #386, opened 2018-01-30
Updated TaskLauncherSink to be able to use SCST 2.0

### LINKED ISSUE #385: TaskLauncherSink needs to support SCSt 2.0
Upgrade TaskLauncherSink to accept messages instead of the TaskLaunchRequest POJO alone.  And still be able to handle POJO and String payload.

cppwfs:
* Updated tests
 - to handle additional cases
 - removed `now` invalid tests
* removed logging payload from the TaskLauncher
* handles message instead TaskLaunchRequestPOJO

resolves #385

### DIFF
--- a/spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
+++ b/spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
```diff
@@ -48,19 +53,48 @@
...
+			Map<String, Object> linkedHashMap = (Map<String, Object>)request.getPayload();
+			taskLaunchRequest = new TaskLaunchRequest((String)linkedHashMap.get("uri"),
+					(List<String>) linkedHashMap.get("commandlineArguments"),
+					(Map)linkedHashMap.get("environmentProperties"),
+					(Map)linkedHashMap.get("deploymentProperties"),
+					(String)linkedHashMap.get("applicationName"));
+		}
+		else {
+			throw new IllegalArgumentException("Unsupported payload type");
+		}
```
mminella: Why do we need to do this type conversion in our code and not have SCSt handle it?

cppwfs: Adding @sobychacko for color commentary.  There maybe a better way to handle this, but it was my understanding that this was the correct method.

mminella: cc: @olegz Reading this code makes me feel like either we're doing it wrong or there is something missing from Spring Cloud Stream.

cppwfs: Let me add a little more commentary.   We looked at using the https://github.com/spring-cloud-stream-app-starters/core/blob/master/common/app-starters-postprocessor-common/src/main/java/org/springframework/cloud/stream/app/postprocessor/ContentTypeEnvironmentPostProcessor.java[ContentTypeEnvironmentPostProcessor] available in the Spring Cloud Stream Apps Starters project, but that didn't translate well for SCT.  We could have implemented for the TaskLauncherStarter, but in this case the solution work for those who wish to use the TaskLauncher outside of the starters project.

sabbyanandan:
@sobychacko: any thoughts on this PR would be helpful

spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLauncherSink.java
sobychacko: @cppwfs I might have missed this context when we chatted. However, since you made the `TaskLaunchRequest` object `json` friendly, I think you can simply rely on the default content type used by spring cloud stream which is `application/json` to do the conversion. It should convert the payload appropriately to the POJO and you can remove all the object mapper based conversions as the the incoming payload is already `TaskLaunchRequest`.

sobychacko: You can force it to use `application/json` as the content type by using this property: `spring.cloud.stream.bindings.input.contentType=application/json`

cppwfs:
@sobychacko sorry about the confusion.  Updated the PR simplifying the ServiceActivator.   Also cleaned up the tests  a little.

mminella:
Rebased, squashed, and merged as 7843ba2edb3d9e9a92b54467fed367986bd0a9a3.

PR #388, opened 2018-02-01
Deprecate the use of spring.cloud.task.batch.listener.enable

### LINKED ISSUE #379: spring.cloud.task.batch.listener.enable is not standard
it should  be `spring.cloud.task.batch.listener.enabled`

cppwfs:
replace it with spring.cloud.task.batch.listener.enabled

resolves #379

### DIFF
--- a/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchAutoConfiguration.java
+++ b/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchAutoConfiguration.java
@@ -32,11 +32,14 @@ import org.springframework.context.annotation.Configuration;
 /**
  * Provides auto configuration for the {@link TaskBatchExecutionListener}.
  *
+ * The spring.cloud.task.batch.listener.enable is being deprecated,
+ * spring.cloud.task.batch.listener.enabled should be used.
+ *
  * @author Michael Minella
  */
 @Configuration
 @ConditionalOnBean({Job.class})
-@ConditionalOnProperty(name = "spring.cloud.task.batch.listener.enable", havingValue = "true", matchIfMissing = true)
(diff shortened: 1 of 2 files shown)


spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/configuration/TaskBatchAutoConfiguration.java
```diff
@@ -32,11 +32,14 @@
 /**
  * Provides auto configuration for the {@link TaskBatchExecutionListener}.
  *
+ * The spring.cloud.task.batch.listener.enable is being deprecated,
```
mminella: `is deprecated`...not `is being deprecated` (makes it sound like we're going to get to it later).

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/listener/TaskBatchExecutionListenerTests.java
```diff
@@ -86,6 +86,36 @@ public void testAutobuiltDataSource() {
...
+	@Test(expected = AssertionError.class)
+	public void testNoAutoConfigurationBothDisabled() {
+		this.applicationContext = SpringApplication.run(new Class[] {JobConfiguration.class,
+				PropertyPlaceholderAutoConfiguration.class,
+				EmbeddedDataSourceConfiguration.class,
+				BatchAutoConfiguration.class,
+				TaskBatchAutoConfiguration.class}, new String[] {"--spring.cloud.task.batch.listener.enable=false --spring.cloud.task.batch.listener.enabled=false"});
+		validateContext();
+	}
+
```
mminella: Can we get a test where the user explicitly sets the property value to true?

cppwfs: Forgot to add comment here.  The PR was updated with that test.

mminella:
LGTM.  Rebase, squashed, and merged as b6f6d68385fbec804248a95e5c48948b03153246

PR #404, opened 2018-03-07
Updated partition integration test to disable feature in deployer-local

### LINKED ISSUE #403: Partition Sample Integration Test fails unless JSON property is set to false
Must add `--spring.cloud.deployer.local.use-spring-application-json=false` to app.run command line args so that they will be passed to the deployer local properly.

cppwfs:
Fixes issue where we have to disable the deployer-local's use of SPRING_APPLICATION_JSON
resolves #403


mminella:
LGTM.  Rebased and merged as b9d95df0ff6882c4f536e9091f6e10c2f628d1f3

PR #433, opened 2018-07-05
Added spring-boot-autoconfigure-processor to the active modules of SCT

### LINKED ISSUE #432: Use Spring Boot autoconfigure-processor to optimize auto-configurations
As explained in [the its reference documentation](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc#autoconfigure-module), Spring Boot provides a `spring-boot-autoconfigure-processor`.

This should be added as an optional dependency to the modules that contains the AutoConfiguration classes (and contains the `spring.factories` file declaring them).

Once configured, this annotation processor will generate metadata at compile time and will help Spring Boot process those auto-configurations without loading the actual classes in some cases.

In this project, several modules are declaring at least one auto-configuration but don't use this annotation processor. Could you add it to the build?

```xml
<dependency>
	<artifactId>spring-boot-autoconfigure-processor</artifactId>
	<optional>true</optional>
</dependency>
```

Once added, this should generate a `spring-autoconfigure-metadata.properties` file in the `target/classes/META-INF` folders.

cppwfs:
resolves #432

### DIFF
--- a/spring-cloud-task-batch/pom.xml
+++ b/spring-cloud-task-batch/pom.xml
@@ -79,5 +79,10 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-test</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-autoconfigure-processor</artifactId>
+			<optional>true</optional>
+		</dependency>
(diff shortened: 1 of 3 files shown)


mminella:
LGTM.  Rebased and merged.

PR #434, opened 2018-07-10
Implement getRunningTaskCount

### LINKED ISSUE #352: TaskLauncherSink needs to provide facility for maximum number of simultaneous tasks executed.
Scenario:  The case of "file ingest" where we will end up with messages on a message queue with (remote) file to be processed by tasks. If there are 100 messages in the queue, we don't want to launch 100 task requests, but limit the number. Seems like some tracking of the tasks launched and a listener for task completion need to be in the task-launcher

dturanski:
Fixes #352

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
@@ -140,6 +140,17 @@ public class MapTaskExecutionDao implements TaskExecutionDao {
 	}
 
 	@Override
+	public long getRunningTaskExecutionCount() {
+		int count = 0;
+		for (Map.Entry<Long, TaskExecution> entry : taskExecutions.entrySet()) {
+			if ( entry.getValue().getEndTime() == null) {
+				count++;
+			}
+		}
+		return count;
+	}
+
+	@Override
 	public long getTaskExecutionCount() {
 		return taskExecutions.size();
 	}
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
@@ -41,6 +41,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Glenn Renfro
  * @author Gunnar Hillert
+ * @author David Turanski
  */
 public class MapTaskExecutionDao implements TaskExecutionDao {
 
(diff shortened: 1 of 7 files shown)


mminella:
Why do we need to add this to Spring Cloud Task?  This feels like something the UI needs in SCDF which isn't directly related to SCT's functionality.

cppwfs:
the findRunningTaskExecutions() does so for a specific task name, not for all running tasks.

cppwfs:
Code LGTM.  Wouldn't we want to expose this method through the TaskExplorer instead just the DAOs?

mminella:
I still don't see why this belongs in SCT.  Also +1 for if you are going to modify the DAOs the `TaskExplorer` interface should be modified.  The DAOs are intended to be only used by the `TaskExplorer` and `TaskRepository` and not directly in the first place.

sabbyanandan:
SCDF or the UI do not directly require it, but it is necessary for the TaskLauncher (via SCDF's API) to decide whether it can kickoff another launch request. You can refer to the parent issue that links to all the downstream stories that depend on it.

Though this doesn't directly add value in SCT, it will help with the integration and the orchestration mechanics of Tasks in SCDF on Cloud Foundry and Kubernetes. In other words, it is a general purpose improvement required for the file-ingest solution architecture on cloud platforms.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
```diff
@@ -139,6 +140,17 @@ public long getRunningTaskExecutionCountByTaskName(String taskName) {
 		return count;
 	}
 
+	@Override
+	public long getRunningTaskExecutionCount() {
+		int count = 0;
```
cppwfs: int needs to be a long

cppwfs: May also consider replacing the for loop with: 
```
count = taskExecutions.entrySet().stream()
	.filter(entry -> entry.getValue().getEndTime() == null)
	.count();
```

cppwfs:
If @mminella is good with this PR.  And @dturanski doesn't mind my changes I can merge.

dturanski:
Fine w me

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/MapTaskExecutionDao.java
mminella: I agree on the `long`, I personally disagree with the changing a for loop to a stream just for syntactic sugar.  I'm personally a fan of not using streams for this use case (it's a performance hog: https://jaxenter.com/java-performance-tutorial-how-fast-are-the-java-8-streams-118830.html).

cppwfs:
Fixed code review comments on merge.  Squashed, merged, backported.

PR #460, opened 2018-10-22
Exit code is nullable

ilayaperumalg:
- Make exit code nullable and only update of execution status can make it a valid integer
 - Update JDBC create/start task execution queries to have exitCode as `null` values
 - Update tests to validate/verify the appropirate exit code values for create/start/complete task executions

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/TaskExecution.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/TaskExecution.java
@@ -27,6 +27,7 @@ import org.springframework.util.Assert;
  *
  * @author Glenn Renfro
  * @author Michael Minella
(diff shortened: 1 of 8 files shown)


cppwfs:
Overall LGTM.  But did find one problem in the TaskLauncher Integration Test.   It sporadically fails because it expects a `0` `exitCode`.   And from time to time the app is isn't finished before the assert is fired.    
If you don't mind I can apply the change (minus the pom.xml change) in this branch to the PR on merge.   https://github.com/cppwfs/spring-cloud-task/commit/7f64302fe5f08193ddd182fc67afcc4c76662781

ilayaperumalg:
@cppwfs Sure, thank you for updating that change!

cppwfs:
LGTM, 
Updated test, squashed merged.

PR #484, opened 2018-11-15
Cleanup code smells reported by Sonar

cppwfs:
This is a first pass are removing some of the code smells.
Many reported code smells were ignored in this effort


markpollack:
rebased and merged as 8e1d38a76af35602bfdbc04ebefe4573a4b8552e

PR #485, opened 2018-11-15
Added discussion on using SmartLifecycle#stop to java docs.

### LINKED ISSUE #471: Docs need to be updated to state that doTaskEnd is now handled by stop() event not destroy()

cppwfs:
Also updated other docs in the TaskLifecycleListener.

resolves #471

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
@@ -60,12 +60,14 @@ import org.springframework.util.StringUtils;
  *     is expected to contain a single application context.</li>
  *     <li>{@link ApplicationReadyEvent} - Used to identify the successful end of a task.</li>
  *     <li>{@link ApplicationFailedEvent} - Used to identify the failure of a task.</li>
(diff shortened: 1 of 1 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -60,12 +60,19 @@
...
+ *     is not emitted. This can occur if an error occurs while executing a BeforeTask.
+ *     </li>
  * </ul>
  *
- * <b>Note:</b> By default, the context will be closed at the completion of a task (once
- * the task repository has been updated).  This behavior can be configured via the
- * property <code>spring.cloud.task.closecontext.enable</code> (defaults to true).
- * Also if the context did not start the FailedTask and TaskEnd may not have all the dependencies met.
+ * <b>Note:</b> By default, the context will not be closed at the completion of
+ * a task if there are daemon threads still running.  This behavior can be configured via the
```
mminella: I'm tweaking the wording of this.  Since 99% of the time, there will not be any daemon threads running, I'd like to word this in the reverse "By default, the context will close at the completion of the task unless other non-daemon threads keep it running".  And it's non-daemon threads...daemon threads do not keep a JVM running.

mminella:
Updated, squashed, and merged as ab58cdd9dffc711f417460d50a0e238efef1bc9b

PR #488, opened 2018-11-29
Rely on spring-cloud-deployer autoconfig

jvalkeal:
- Deployer core/local uses 2.0.0.BUILD-SNAPSHOT
- Use pring-cloud-deployer-dependencies bom
- Remove ResourceLoadingAutoConfiguration as that is now in
  deployer itself with a way to configure loaders.
- spring-cloud-task-core now depends on spring-cloud-deployer-autoconfigure.
- Rest of deployer modules like support/maven are still defined in deps
  as those are optional in spring-cloud-deployer-autoconfigure.


spring-cloud-task-core/pom.xml
```diff
@@ -19,6 +19,10 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-autoconfigure</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-deployer-autoconfigure</artifactId>
```
mminella: Why is this needed in core?  I'd expect this to be needed in the stream module...

jvalkeal: Yeah, I wasn't sure which one needed this and just picked core as it already have deps for `spring-cloud-deployer-resource-support` and `spring-cloud-deployer-resource-maven`.

cppwfs: It is needed in both the stream and batch module.  The batch uses it for partitioning.

jvalkeal: ah, I think that `ObjectProvider.orderedStream()` were added to framework 5.1.x. I suppose if task builds on boot21/framework51, you would not expect boot20 apps be fully supported.

jvalkeal: task autoconfig is currently in `spring-cloud-task-core` so technically it's in a same place/level now with this pr.

cppwfs: My bad on the first test.   I got it working.
LGTM

PR #492, opened 2018-12-03
Add documentation on when to resolve SCT delcared beans.

### LINKED ISSUE #491: Add documentation on when to resolve task specific beans
In the previous releases of SCT we used an import in the @EnableTask annotation.   So in cases that a user used a SCT declared bean `TaskExplorer` they could resolve it in their `@Configuration` classes.   Now that we are using AutoConfigure, this will no longer work.  They need to resolve it in class that needs it.

cppwfs:
resolves #491

### DIFF
--- a/spring-cloud-task-docs/src/main/asciidoc/features.adoc
+++ b/spring-cloud-task-docs/src/main/asciidoc/features.adoc
@@ -233,6 +233,10 @@ custom implementation of the `TaskConfigurer` interface. Typically, extending th
 overriding the required getter is sufficient. However, implementing your own from scratch
 may be required.
 
+NOTE: Users should not directly use getter methods from a TaskConfigurer directly
+unless they are using it to supply beans in a class annotated with
+https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Configuration.html[Configuration].
+
 [[features-task-name]]
 === Task Name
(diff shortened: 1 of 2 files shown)


mminella:
I'm not clear what this PR actually says.  Based on our previous conversations, I'd expect us to do two things:

1. Update the javadoc on the `TaskConfigurer` to be clear that the resulting beans should not be directly referenced unless the provided configuration is not used.
2. Update the reference docs when introducing the point about customizing the `TaskConfigurer` to point out the same detail.

spring-cloud-task-docs/src/main/asciidoc/features.adoc
```diff
@@ -233,6 +233,10 @@ custom implementation of the `TaskConfigurer` interface. Typically, extending th
 overriding the required getter is sufficient. However, implementing your own from scratch
 may be required.
 
+NOTE: Users should not directly use getter methods from a TaskConfigurer directly
+unless they are using it to supply beans in a class annotated with
+https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Configuration.html[Configuration].
```
mminella: I'll tweak this on merge since technically you could use XML configuration as well...

mminella:
Rebased, squashed, and merged.

PR #501, opened 2019-01-03
Updated the DB2 Script to fix the not null in primary key.

### LINKED ISSUE #500: Broken db2 sql for TASK_LOCK
Primary key columns need to be defined as not null as shown below:

```
db2 => CREATE TABLE TASK_LOCK (LOCK_KEY CHAR(36), REGION VARCHAR(100), CLIENT_ID CHAR(36), CREATED_DATE TIMESTAMP NOT NULL, constraint LOCK_PK primary key (LOCK_KEY, REGION))
DB21034E  The command was processed as an SQL statement because it was not a 
valid Command Line Processor command.  During SQL processing it returned:
SQL0542N  The column named "LOCK_KEY" cannot be a column of a primary key or 
unique key constraint because it can contain null values.  SQLSTATE=42831

db2 => CREATE TABLE TASK_LOCK (LOCK_KEY CHAR(36) NOT NULL, REGION VARCHAR(100) NOT NULL, CLIENT_ID CHAR(36), CREATED_DATE TIMESTAMP NOT NULL, constraint LOCK_PK primary key (LOCK_KEY, REGION))
DB20000I  The SQL command completed successfully.
```

cppwfs:
resolves #500


cppwfs:
@jvalkeal thanks for finding and reporting this.

mminella:
Why is this required for us and not for SI?  https://github.com/spring-projects/spring-integration/blob/master/spring-integration-jdbc/src/main/resources/org/springframework/integration/jdbc/schema-db2.sql

cppwfs:
It fails as is, based on my tests.   I will open a issue and PR on SI.

mminella:
@garyrussell or @artembilan We've taken the schema from Spring Integration's JDBC lock.  Have you had any issues with this?

artembilan:
No, we don't. Just because our client for that table never allows to pass nulls:
```
// For REGION
public void setRegion(String region) {
	Assert.hasText(region, "Region must not be null or empty.");
	this.region = region;
}
...

// For LOCK_KEY
private String pathFor(String input) {
	return input == null ? null : UUIDConverter.getUUID(input).toString();
}
```

But we are pretty open to add such a fix into SI as well.

Thanks

mminella:
@cppwfs Aren't we using the SI client for those tables?

cppwfs:
Looking at the SI implementation you will get alot of tables we don't need.

mminella:
LGTM.  Rebased and merged.

PR #506, opened 2019-01-16
Updated all task_lock tables to add NOT NULL to primary keys.

### LINKED ISSUE #505: Need to add not null to all task_lock primary keys
This is to match the DDL that is used by Spring Integration for the lock table.

cppwfs:
resolves #505


mminella:
LGTM.  Rebased, merged, and backported.

PR #517, opened 2019-02-21
Remove Spring Batch Versions

mminella:
Since Spring Cloud Task depends on Spring Boot, it should obtain the
version of dependencies managed by Spring Boot from Spring Boot.  This
commit removes the explict configuration of Spring Batch's version in
the Spring Cloud Task pom and instead delegates that to Spring Boot.


cppwfs:
Rebased, polish, merge

PR #519, opened 2019-02-21
Updated Base Version to 2.2.x

cppwfs:
Removed Task Dependencies.  These should be brought in by Boot


cppwfs:
Closing as the branch name was incorrect.   Using TASK-518

PR #521, opened 2019-02-22
Task baseline 2.1.1

cppwfs:


spring-cloud-task-core/src/test/java/org/springframework/cloud/task/repository/support/DatabaseTypeTests.java
```diff
@@ -39,10 +40,11 @@
...
 	public void testFromProductName() {
-		assertEquals(HSQL, fromProductName("HSQL Database Engine"));
-		assertEquals(ORACLE, fromProductName("Oracle"));
-		assertEquals(POSTGRES, fromProductName("PostgreSQL"));
-		assertEquals(MYSQL, fromProductName("MySQL"));
+		assertThat(fromProductName("HSQL Database Engine")).isEqualTo(HSQL);
+		assertThat(fromProductName("Oracle")).isEqualTo(ORACLE);
+		assertThat(fromProductName("PostgreSQL")).isEqualTo(POSTGRES);
+		assertThat(fromProductName("MySQL")).isEqualTo(MYSQL);
+		assertThat(fromProductName("MariaDB")).isEqualTo(MYSQL);
```
mminella: We should leave these as assertEquals and just add the MariaDB check since we are not making the global change to assertj on 2.1.x

PR #562, opened 2019-03-25
Fix closecontextEnabled property reference

### LINKED ISSUE #561: Incorrect reference to spring.cloud.task.closecontext.enabled
According to https://github.com/spring-cloud/spring-cloud-task/blob/2bff013b83e0fa084616b0de0f3fc3767ccfd435/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java#L70 the property is `spring.cloud.task.closecontext.enabled` however, this does not bind properly.

Instead, it should be `spring.cloud.task.closecontextEnabled` or `spring.cloud.task.closecontext_enabled`.

donovanmuller:
As per reference used in [docs](https://github.com/spring-cloud/spring-cloud-task/blob/master/spring-cloud-task-docs/src/main/asciidoc/features.adoc).

Fixes #561

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
@@ -67,8 +67,8 @@ import org.springframework.util.StringUtils;
  *
  * <b>Note:</b> By default, the context will close at the completion of the task unless
  * other non-daemon threads keep it running. Programatic closing of the context can be
- * configured via the property <code>spring.cloud.task.closecontext.enabled</code>
- * (defaults to false). If the <code>spring.cloud.task.closecontext.enabled</code> is set
+ * configured via the property <code>spring.cloud.task.closecontext_enabled</code>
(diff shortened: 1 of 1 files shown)


mminella:
We actually have unit tests that use both syntaxes within the framework ([1][2]).  I believe that since the dot (`.`) notation is preferred unless setting these via environment vars, we probably should leave this as is.  I'd actually recommend, instead, to note in the docs that either syntax is supported...

[1] https://github.com/spring-cloud/spring-cloud-task/blob/master/spring-cloud-task-core/src/test/java/org/springframework/cloud/task/TaskCoreTests.java#L76
[2] https://github.com/spring-cloud/spring-cloud-task/blob/master/spring-cloud-task-core/src/test/java/org/springframework/cloud/task/listener/TaskLifecycleListenerTests.java#L170

donovanmuller:
I'll have to check again but the only reason for raising this issue is that the dot notation did not work at all in practice.

I'll feedback once I've verified.

cppwfs:
Rebased, Merged.

Donovan, 
Thanks for the contribution!!!!!

PR #598, opened 2019-05-09
Initialize values with properties vs value

### LINKED ISSUE #587: Properties used buy TaskRepositoryInitializer should be in TaskProperties
Currently `spring.cloud.task.initialize.enable` and `spring.cloud.task.tablePrefix` are only referenced via the `@Value`  in the TaskRepositoryInitializer.  Thus when a user tries to access them via properties in their IDE, the IDE marks them as not found/incorrect.   Refer to issue https://github.com/spring-cloud/spring-cloud-task/issues/584

cppwfs:
resolves #587


mminella:
Why are we creating a new Boot Properties object for a single boolean flag?  Can't we just add this to the existing `TaskProperties`?

cppwfs:
Good question.   I had to look it up.  But it was so that we could keep the `initialize` portion of the `spring.cloud.task.initialize.enable` property.

mminella:
I would just depricate the property and use a new one under the other class.

mminella:
Am I missing something or do we not have a test for the old property as well?  If the old property is depricated, we still need to support it until we remove it.

cppwfs:
Re-add the tests...  But I wanted to go to Toshi station and pick up power converters.

mminella:
Rebased, squashed, and merged.

PR #607, opened 2019-06-14
Updated the order in which the TaskLifecycleListener is configured

### LINKED ISSUE #605: Task status is never set to KO when I use a bootstrap file and an exception is thrown in a task
I'm working on the Greenwich RELEASE version of Spring Cloud.

I have created a Task connected to my config server and set spring.cloud.task.batch.fail-on-job-failure to true.

When an exception is thrown in my Task, in case I use a bootstrap file, the status of the Task is never in error. I have the right status when I remove the connection to the config server and add an application file.

After investigation, here's what I think I understood: in the `BootstrapApplicationListener` class, a new `CloseContextOnFailureApplicationListener` listener has been added. This listener supports the `ApplicationFailedEvent` event. To put the status of the Task in error, the `TaskLifecycleListener` listener is also based on the `ApplicationFailedEvent` event. Unfortunately, `CloseContextOnFailureApplicationListener` processes the `ApplicationFailedEvent` event first and returns a `ContextClosedEvent` event via the `context.close()` method. This event ends the Task but via the method `TaskLifecycleListener.stop()` without updating the fact that the task is in error.

Afterwards, it is finally the `TaskLifecycleListener` that supports the `ApplicationFailedEvent` event but as the Task is already finished, the task status is not updated.

How can I run a Task with a configuration server and have the right status (OK, KO)? Can I change the order of the listeners so that the `ApplicationFailedEvent` event is first processed by `TaskLifecycleListener`?

cppwfs:
resolves #605

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
@@ -45,6 +45,7 @@ import org.springframework.context.ApplicationEvent;
 import org.springframework.context.ApplicationListener;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.SmartLifecycle;
+import org.springframework.core.Ordered;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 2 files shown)


mminella:
This PR probably needs documentation added to it to indicate the risk of trying to place a listener earlier in the lifecycle.

mminella:
Polished, merged, and backported to 2.1.y and 2.0.x.

PR #609, opened 2019-06-25
Allows users to set spring.batch.job.enabled with fail-on-job-failure.

### LINKED ISSUE #525: Task app launching/restarting batch-jobs depending on properties (conflict with Enable task.batch.failOnJobFailure )
Hi,

i want to be able to restart or launch a job depending on properties.
I achieved this by writing my own CommandLineRunner that uses a jobOperator.

A sample application would be one that processes files.
With three properties : `batch.jobname`, `batch.filename` and `batch.restart`
So I can launch a task with properties : job1, file1, false
If I then see the job failed, I can launch the same task with properties: job1, file1, true
(This is useful, if i know another service that was called in the processing was down and is up now again)
If batch.restart == true, I lookup the lastExecution of job1 and restart via the JobOperator.
And if batch.restart == false, I just launch the nextInstance.

To disable autostart of jobs I set `spring.batch.job.enabled: false` 
As I also wanted my task to fail if any of the jobs fail I set `spring.cloud.task.batch.failOnJobFailure: true` aswell

But this annotationn makes use of [tasks own CommandLineRunner](
https://github.com/spring-cloud/spring-cloud-task/blob/master/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/handler/TaskJobLauncherCommandLineRunner.java) that will just launch all available jobs.

[Here](https://gitlab.com/jmundorf/spring-cloud-task-timestamp-batch) is an example of the launch/restart functionality for the timestamp example.

So I think I either need another approach to separate launching and restarting or override the TaskJobLauncherCommandLineRunner.java ?

cppwfs:
resolves #525

### DIFF
--- a/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/handler/TaskJobLauncherCommandLineRunnerTests.java
+++ b/spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/handler/TaskJobLauncherCommandLineRunnerTests.java
@@ -32,6 +32,7 @@ import org.springframework.batch.core.configuration.annotation.DefaultBatchConfi
 import org.springframework.batch.core.configuration.annotation.EnableBatchProcessing;
 import org.springframework.batch.core.configuration.annotation.JobBuilderFactory;
 import org.springframework.batch.core.configuration.annotation.StepBuilderFactory;
+import org.springframework.batch.core.explore.JobExplorer;
 import org.springframework.batch.core.launch.JobLauncher;
 import org.springframework.batch.core.launch.support.SimpleJobLauncher;
 import org.springframework.batch.core.scope.context.ChunkContext;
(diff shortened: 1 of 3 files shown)


spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/handler/TaskJobLauncherCommandLineRunnerTests.java
```diff
@@ -112,6 +112,17 @@ public void testTaskJobLauncherCLRSuccessFailWithTaskExecutor() {
...
+	@Test
+	public void testNoTaskJobLauncher() {
+		String[] enabledArgs = new String[] {
+				"--spring.cloud.task.batch.failOnJobFailure=true",
+				"--spring.cloud.task.batch.failOnJobFailurePollInterval=500",
+				"--spring.batch.job.enabled=false" };
+		this.applicationContext = SpringApplication.run(new Class[] {
+				TaskJobLauncherCommandLineRunnerTests.JobWithFailureConfiguration.class },
+				enabledArgs);
+	}
```
mminella: What does this verify?  There are no assertions.

cppwfs: The app that is executed has 2 jobs and one of them throws an exception.   So in this case we know the app did not execute because there is no exception   thrown.

cppwfs: get jobRepository from context and check the jobexecution count.

mminella:
Rebased, squashed, merged, and backported.

PR #621, opened 2019-07-29
Capture external-execution-id for child tasks started from a partition

### LINKED ISSUE #620: Task apps launched using partitioning should have external-execution-id populated.

cppwfs:
resolves #620

### DIFF
--- a/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
+++ b/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
@@ -74,6 +74,7 @@ import org.springframework.util.StringUtils;
  * </p>
  *
  * @author Michael Minella
+ * @author Glenn Renfro
  */
 public class DeployerPartitionHandler
 		implements PartitionHandler, EnvironmentAware, InitializingBean {
@@ -97,6 +98,11 @@ public class DeployerPartitionHandler
 	 * ID of Spring Cloud Task parent execution.
 	 */
 	public static final String SPRING_CLOUD_TASK_PARENT_EXECUTION_ID = "spring.cloud.task.parentExecutionId";
+
+	/**
+	 * ID of the Spring Cloud Task execution.
+	 */
+	public static final String SPRING_CLOUD_TASK_EXECUTION_ID = "spring.cloud.task.executionid";
 
 	/**
 	 * Spring Cloud Task name property.
@@ -142,6 +148,16 @@ public class DeployerPartitionHandler
 	@Autowired
 	private TaskRepository taskRepository;
 
+	/**
+	 * Constructor initializing the DeployerPartitionHandler instance.
+	 * @param taskLauncher the launcher used to execute partitioned tasks.
+	 * @param jobExplorer used to acquire the status of the job.
+	 * @param resource the url to the app to be launched.
+	 * @param stepName the name of the step.
+	 * @deprecated Use the constructor that accepts {@link TaskRepository} as well as the
+	 * this constructor's set of parameters.
+	 */
+	@Deprecated
 	public DeployerPartitionHandler(TaskLauncher taskLauncher, JobExplorer jobExplorer,
 			Resource resource, String stepName) {
 		Assert.notNull(taskLauncher, "A taskLauncher is required");
@@ -161,7 +177,7 @@ public class DeployerPartitionHandler
 		Assert.notNull(jobExplorer, "A jobExplorer is required");
 		Assert.notNull(resource, "A resource is required");
 		Assert.hasText(stepName, "A step name is required");
-		Assert.notNull(taskRepository, "TaskRepository must not be null");
+		Assert.notNull(taskRepository, "A TaskRepository is required");
 
 		this.taskLauncher = taskLauncher;
 		this.jobExplorer = jobExplorer;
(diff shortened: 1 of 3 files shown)


mminella:
Overall looks good.  My only question is if there is a way to handle the added constructor param in a non-breaking way.  Thoughts?

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
```diff
@@ -919,14 +927,18 @@ public int compare(AppDeploymentRequest o1, AppDeploymentRequest o2) {
 			assertThat(request.getCommandlineArguments().contains(formatArgs(
 					DeployerPartitionHandler.SPRING_CLOUD_TASK_STEP_NAME, "step1")))
 							.isTrue();
+			request.getCommandlineArguments().forEach(arg -> System.out.println("****"+arg));
```
cppwfs: OOOPs... This one slipped past.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
```diff
@@ -329,11 +349,22 @@ private void launchWorker(StepExecution workerStepExecution) {
...
-				this.deploymentProperties, arguments);
+				this.deploymentProperties, commandLineArgsPrep(taskExecution, arguments));
+
+		String externalExecutionId = this.taskLauncher.launch(request);
+		this.taskRepository.updateExternalExecutionId(taskExecution.getExecutionId(),
+				externalExecutionId);
+	}
 
-		this.taskLauncher.launch(request);
+	private List<String> commandLineArgsPrep(TaskExecution taskExecution,
```
mminella: We let the user choose if the args we provide are passed via command line args or env vars.  This property should be treated the same.  See lines 313 and 334.

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
```diff
@@ -79,14 +80,17 @@
...
 	@Before
 	public void setUp() {
 		MockitoAnnotations.initMocks(this);
 		this.environment = new MockEnvironment();
+		TaskExecution taskExecution = new TaskExecution(2, 0, "name", new Date(),
+				new Date(), "", Collections.emptyList(), null, null, null);
+		when(taskRepository.createTaskExecution()).thenReturn(taskExecution);
 	}
 
 	@Test
```
mminella: We are only doing a test on the old constructor's validation.  We need one for the new constructor as well, especially since the new one does not use the old one (although it could).

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
```diff
@@ -60,6 +60,7 @@
 
 /**
  * @author Michael Minella
+ * @author Glenn Renfro
  */
 public class DeployerPartitionHandlerTests {
```
mminella: Based on my earlier comment about giving the user the option to pass in the id via env vars of command line args, we'll need a test for the env var option as well.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
```diff
@@ -137,6 +139,9 @@
 
 	private boolean defaultArgsAsEnvironmentVars = false;
 
+	@Autowired
+	private TaskRepository taskRepository;
+
 	public DeployerPartitionHandler(TaskLauncher taskLauncher, JobExplorer jobExplorer,
```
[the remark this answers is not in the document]
cppwfs: Deprecated

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
cppwfs: Done

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
cppwfs: Done

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
```diff
@@ -290,6 +326,8 @@ private void launchWorker(StepExecution workerStepExecution) {
 
 		arguments.addAll(this.commandLineArgsProvider.getCommandLineArgs(copyContext));
 
+		TaskExecution partitionTaskExecution = this.taskRepository.createTaskExecution();
```
mminella: How does this work if the old constructor is used?

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
```diff
@@ -333,7 +375,9 @@ private void launchWorker(StepExecution workerStepExecution) {
 		AppDeploymentRequest request = new AppDeploymentRequest(definition, this.resource,
 				this.deploymentProperties, arguments);
 
-		this.taskLauncher.launch(request);
+		String externalExecutionId = this.taskLauncher.launch(request);
+		this.taskRepository.updateExternalExecutionId(
+				partitionTaskExecution.getExecutionId(), externalExecutionId);
```
mminella: Same comment here about the missing task repository

spring-cloud-task-batch/src/test/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandlerTests.java
mminella: The new test added only tests it's validation...not that it still works with the new code (see my comments above).

mminella:
Rebased, polished, squashed, and merged.

PR #626, opened 2019-08-22
Resolve issue where spring.batch.job.names was not expressed

### LINKED ISSUE #614: TaskJobLauncherCommandlineRunner should use boot properties
Spring boot offers the ability to select which jobs are run for a Spring Batch job, by allowing a user to set the job names on the ` spring.batch.job.names ` property.   When Spring Cloud Task uses the TaskJobLauncherCommandlineRunner it uses the `spring.cloud.task.batch.jobNames` property.   We should match the boot property including `order`.

cppwfs:
when using TaskJobLauncherCommandLineRunner

After research it was determined the only affected property was `spring.batch.job.names`.   Also deprecated `spring.cloud.task.batch.jobNames`
resolves #614


mminella:
Merged and polished.

PR #630, opened 2019-10-14
Updated the test so that it will now work.

cppwfs:
In order to execute the test do a full `mvn clean install`.   Then execute a `mvn clean package -P integrationTests -pl spring-cloud-task-samples/partitioned-batch-job/`

We will also need to add a new task to the bamboo integration tests for SCT.
In this task we will pass in the parameters `-P integrationTests -pl spring-cloud-task-samples/partitioned-batch-job/`
This will be run after the other integration tasks.


mminella:
Rebased and merged.

PR #639, opened 2019-12-05
Updated diagram

### LINKED ISSUE #627: Task Repository Schema not uptodate
The ERD shown at https://docs.spring.io/spring-cloud-task/docs/2.1.1.RELEASE/reference/htmlsingle/#appendix-task-repository-schema is not uptodate. The following columns are missing:

- `TASK_EXECUTION.external_execution_id`
- `TASK_EXECUTION.parent_execution_id`

Also, the following things are not documented:

- indices
- initial values (e.g. in `TASK_SEQ`)

It would be very helpful if you could document the whole DDL and not just show the ERD. The reasons are the same as I reported for SCDF at https://github.com/spring-cloud/spring-cloud-dataflow/issues/3460

In short there always are situations where one needs to create the DB objects by hand, in our case we need to create each table with the correct `ROW_FORMAT` defined. If we don't do so, the creation of indices fails with `Index column size too large. The maximum column size is 767 bytes.`. The whole problem is also documented here: https://community.pivotal.io/s/article/Apps-are-down-due-to-the-Maximum-Column-Size-is-767-bytes-Constraint-in-MySQL

For sure there are other situations where it is required to create the tables in some specific way.

cppwfs:
resolves #627


spring-cloud-task-docs/src/main/asciidoc/appendix-task-repository-schema.adoc
```diff
@@ -7,3 +7,79 @@ This appendix provides an ERD for the database schema used in the task repositor
...
+|Column Name |Required |Type |Field Length | Default Values |Notes
+
+|TASK_EXECUTION_ID |TRUE |BIGINT | X | X |
+Spring Cloud Task Framework at app startup establishes the next available id as obtained from the `TASK_SEQ`. Or if the record is created outside of task then the value must be populated at record creation time.
+
+|START_TIME |FALSE | DATETIME | X | NULL | Spring Cloud Task Framework at app startup establishes the value.
+
+|END_TIME |FALSE | DATETIME | X | NULL | Spring Cloud Task Framework at app exit establishes the value.
+
+|TASK_NAME |FALSE | VARCHAR | 100 | "" | Spring Cloud Task Framework at app startup will set this to "Application" unless user establish the name using the spring.cloud.task. name as discussed <<features-task-name,here>>
```
mminella: Format `spring.cloud.task.name` (also remove space between `task.` and `name`).

spring-cloud-task-docs/src/main/asciidoc/appendix-task-repository-schema.adoc
```diff
@@ -7,3 +7,79 @@ This appendix provides an ERD for the database schema used in the task repositor
...
+
+|EXIT_CODE |FALSE | INTEGER | X | 0 | Spring Cloud Task Framework at app exit establishes the value.
+
+|EXIT_MESSAGE |FALSE | VARCHAR | 2500 | "" | Spring Cloud Task Framework at app exit establishes the value.
```
mminella: This is user defined: https://docs.spring.io/spring-cloud-task/docs/2.0.0.RELEASE/reference/htmlsingle/#features-task-execution-listener-exit-messages.

mminella:
Squashed, rebased, merged, and backported as d1827eab258a3c89a4199f10f003059bb8765fe3

PR #642, opened 2019-12-20
Resolves Task Lock issue with Oracle DB

cppwfs:
In short the Oracle CHAR type is blank padded to the end of the field.  So we were not getting an exact match on queries when looking for the LOCK_KEY or CLIENT_ID.  
In our release blog we will need to note that Oracle folks will need to run this if upgrading.


mminella:
Rebased and merged both into master and 2.1.y.

PR #644, opened 2020-01-17
fix link to DDL scripts

imod:
fixes the link to the DDL scripts


mminella:
Rebased, merged, and backported.  Thanks for the contribution!

PR #648, opened 2020-02-10
Added section on closing the context

### LINKED ISSUE #646: Document closecontextEnabled flag
Our current reference documentation does not document the `closecontextEnabled` property.

cppwfs:
resolves #646

### DIFF
--- a/spring-cloud-task-docs/src/main/asciidoc/features.adoc
+++ b/spring-cloud-task-docs/src/main/asciidoc/features.adoc
@@ -394,4 +394,16 @@ This can be done either by adding the following annotation to your Task applicat
 ```
 You may also disable Task auto configuration by setting the `spring.cloud.task.autoconfiguration.enabled` property to `false`.
 
+[[closing-the-context]]
+=== Closing the Context
+If the application requires the `ApplicationContext` to be closed at the
+completion of a task (all `*Runner#run` methods have been called and the task
+repository has been updated), set the property `spring.cloud.task.closecontext_enabled`
+to `true`.
 
+Another case to close the context is when the Task Execution completes however the application does not terminate.
+In these cases the context is held open because a thread has been allocated
+(for example: if you are using a TaskExecutor). In these cases
+set the `spring.cloud.task.closecontext_enabled` property to `true` when launching your task.
+This will close the application's context once the task is complete.
+Thus allowing the application to terminate.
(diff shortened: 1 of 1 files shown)


spring-cloud-task-docs/src/main/asciidoc/features.adoc
```diff
@@ -394,4 +394,16 @@ This can be done either by adding the following annotation to your Task applicat
 ```
 You may also disable Task auto configuration by setting the `spring.cloud.task.autoconfiguration.enabled` property to `false`.
 
-
+[[closing-the-context]]
+=== Closing the Context
+If the application requires the `ApplicationContext` to be closed at the
+completion of a task (all `*Runner#run` methods have been called and the task
+repository has been updated), set the property `spring.cloud.task.closecontext_enabled`
```
ilayaperumalg: I am sure there must have been a discussion around naming this property. But, at first look, this property name especially the underscore in `closecontext_enabled` is confusing. Could you help me understand the reasoning of this naming?

cppwfs: I can change it to closeContextEnabled.   In this case I was following the previous usage.   
Would that be helpful?

spring-cloud-task-docs/src/main/asciidoc/features.adoc
```diff
@@ -52,7 +52,7 @@ updated in the repository with the results.
 
 NOTE: If the application requires the `ApplicationContext` to be closed at the
 completion of a task (all `*Runner#run` methods have been called and the task
-repository has been updated), set the property `spring.cloud.task.closecontext_enabled`
+repository has been updated), set the property `spring.cloud.task.closecontextEnabled`
```
ilayaperumalg: The name sounds ok to me. But, where is this implemented? Shouldn't that change there as well?

cppwfs: In this case the property already has the following format `closecontextEnabled`.   i.e. https://github.com/spring-cloud/spring-cloud-task/blob/master/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskProperties.java#L68
I guess we can either use camel case or kabob case.

mminella:
Rebased and merged as f7c250131484362cd489569b6a4e210aa3aee273

PR #660, opened 2020-04-13
Additional docs on table creation when using the tablePrefix prop

### LINKED ISSUE #653: spring.cloud.task.tablePrefix=yourPrefix does not create Tables on its own
@gargpune commented on [Mon Mar 23 2020](https://github.com/spring-cloud/spring-cloud-dataflow/issues/3844)

As per the instructions, I was trying to change the prefix for Spring Cloud Task Tables using property spring.cloud.task.tablePrefix. But this does not create tables on its own with a custom prefix. However, there are DB Specific SQL files. Am I missing anything or in case of a custom prefix, we need to create tables explicitly and to refer those we have to use "spring.cloud.task.tablePrefix".

https://docs.spring.io/spring-cloud-task/docs/current/reference/#features-table-prefix

**Steps to reproduce:**
- Follow hello world example
- Add this property in application.properties spring.cloud.task.tablePrefix=TEST_

**Screenshots:**
Screen Shot 2020-03-23 at 1 51 20 PM

**Additional context:**
If we need to create tables explicitly, Could you please update the documentation ?

cppwfs:
Resolves #653

### DIFF
--- a/spring-cloud-task-docs/src/main/asciidoc/features.adoc
+++ b/spring-cloud-task-docs/src/main/asciidoc/features.adoc
@@ -149,6 +149,13 @@ by setting the `spring.cloud.task.tablePrefix` to the prefix you need, as follow
 
 `spring.cloud.task.tablePrefix=yourPrefix`
 
+By using the `spring.cloud.task.tablePrefix`, a user assumes the responsibility to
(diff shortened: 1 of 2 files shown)


mminella:
Rebased and merged as d508285f58c14fc42625925de19c7fd90bf8b4c7

PR #667, opened 2020-05-11
Add new repo method for task execution Ids retrieval

### LINKED ISSUE #666: Add new repository method for task execution Ids by task name
This is needed for https://github.com/spring-cloud/spring-cloud-dataflow/issues/3903

ilayaperumalg:
- It seems convenient to have a new repo method to get the set of task execution Ids by the given task name.
   This also simplifies multiple repository method invocations/further processing to achieve the same.

 - Add `getTaskExecutionIdsByTaskName` method into TaskExplorer and TaskExecutionDao

 - Implement the new methods and udpate the tests

Resolves #666

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/TaskExplorer.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/TaskExplorer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


ilayaperumalg:
My initial attempt was not to make any changes on SCT but I realized it started making some expensive calls (if a task has huge volume of task executions). My concern is, let’s say we are deleting a task and we also want to delete the associated task executions. The way current setup works is to pass on the execution Ids of the task execution to be cleaned up. If a task has huge number of task executions, then we would have to load all those executions on the REST response (since we can’t go via paging in this case) which I think is heavy. To avoid this, since we only need the task execution Ids, I added this endpoint.

ilayaperumalg:
Closing this with the following discussion outcome:

We'll have the DAO method only in SCDF to take care of the deletions and that way we won't force SCT's TaskExplorer and DAOs implement the new one.

PR #670, opened 2020-06-03
html files now under html dir in zip file

### LINKED ISSUE #668: Unify the docs in spring cloud task
All the other spring cloud projects have a `reference/html/index.html` page that points to the entry for the docs. For task it's `reference/index.html`

cppwfs:
pdf file is now under pdf dir in zip file

resolves #668


mminella:
Our home page for our docs refers to an epub artifact that is not in the zip. If we are creating it, we should link to it, if not, we should remove that link.

mminella:
Epub format was added and pushed.

PR #673, opened 2020-06-09
Added missing `@EnableTask` annotation

FWinkler79:
Following the documentation (using Spring Boot 2.3.0 and Spring Cloud Hoxton.SR5) does not yield the described output if that annotation is missing. I saw it being used in several videos of M. Minella, so I think it is a glitch.
Adding the annotation to my code showed the log output where `TaskExecution` is created and updated in the repository. Without the annotation that does not show (and probably does not happen either).


cppwfs:
LGTM 
Thank you for the contribution!

cppwfs:
Thank you for the contribution!

PR #674, opened 2020-06-09
Enabling custom `TaskNameResolver` implementation

FWinkler79:
The [Spring Cloud Task documentation](https://docs.spring.io/spring-cloud-task/docs/current/reference/#features-task-name) suggests that an application can influence the naming of tasks - by virtue of configuration but also by providing its own implementation of `TaskNameResolver`. The latter is not the case, since `SimpleTaskAutoConfiguration` declares the `TaskNameResolver` bean without `@ConditionalOnMissingBean`.


mminella:
Can you please add a test for this?

cppwfs:
Closing due to inactivity.

PR #684, opened 2020-07-13
Introducing KafkaItemReader for single step jobs

### LINKED ISSUE #679: Create KafkaItemReader Single-Step batch job

cppwfs:
resolves #679
The base set of properties to setup are the following:
```
spring.batch.job.kafkaitemreader.topic=<your topic name>
spring.kafka.consumer.bootstrap-servers=localhost:9092
```

In the case that the string data in the topic needs to be parsed because of inconsistent headers etc
In your app add the following class:
```
package io.spring.kafkasinglestepjob;

import java.util.Map;

import com.fasterxml.jackson.core.JsonProcessingException;
import com.fasterxml.jackson.databind.ObjectMapper;

public class ParseMe {
	public static Object parseMe(String val) {
		ObjectMapper mapper = new ObjectMapper();
		Map result = null;
		try {
			result = mapper.readValue(val, Map.class);
		}
		catch (JsonProcessingException e) {
			e.printStackTrace();
		}
		return result;
	}
}
```
* the parseMe method can have a `String` or a `String, Headers` as a parameter.  
* the code can parse the data as need.   
To use the sample parser above set the following properties:
```
spring.kafka.consumer.value-deserializer=org.springframework.kafka.support.serializer.ParseStringDeserializer
spring.kafka.consumer.properties.spring.message.value.parser=io.spring.kafkasinglestepjob.ParseMe.parseMe
```

NOTE: the default for boot is `ByteArrayDeserializer`.

### DIFF
--- a/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemReaderAutoConfiguration.java
+++ b/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemReaderAutoConfiguration.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2020-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.task.batch.autoconfigure.kafka;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.Properties;
+
+import org.springframework.batch.item.kafka.KafkaItemReader;
+import org.springframework.batch.item.kafka.builder.KafkaItemReaderBuilder;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.batch.BatchAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.kafka.KafkaProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.util.StringUtils;
+
+
+/**
+ *
+ * Autconfiguration for a {@code KafkaItemReader}.
+ * @author Glenn Renfro
+ * @since 2.3
+ */
+@Configuration
+@EnableConfigurationProperties({ KafkaProperties.class, KafkaItemReaderProperties.class })
+@AutoConfigureAfter(BatchAutoConfiguration.class)
+public class KafkaItemReaderAutoConfiguration {
+
+	@Autowired
+	private KafkaProperties kafkaProperties;
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty(prefix = "spring.batch.job.kafkaitemreader", name = "name")
+	public KafkaItemReader<Object, Map<Object, Object>> kafkaItemReader(
+			KafkaItemReaderProperties kafkaItemReaderProperties) {
+		Properties consumerProperties = new Properties();
+		consumerProperties.putAll(this.kafkaProperties.getConsumer().buildProperties());
+		validateProperties(kafkaItemReaderProperties);
+		if (kafkaItemReaderProperties.getPartitions() == null
+				|| kafkaItemReaderProperties.getPartitions().size() == 0) {
+			kafkaItemReaderProperties.setPartitions(new ArrayList<>(1));
+			kafkaItemReaderProperties.getPartitions().add(0);
+		}
+		return new KafkaItemReaderBuilder<Object, Map<Object, Object>>()
+				.partitions(kafkaItemReaderProperties.getPartitions())
+				.consumerProperties(consumerProperties)
+				.name(kafkaItemReaderProperties.getName())
+				.pollTimeout(Duration
+						.ofSeconds(kafkaItemReaderProperties.getPollTimeOutInSeconds()))
+				.saveState(true)
+				.topic(kafkaItemReaderProperties.getTopic())
+				.build();
+	}
+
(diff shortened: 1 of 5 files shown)


cppwfs:
For discussion:  When I put straight text into the topic, the conversion to map failed.   Is this a restriction that we need to enforce?   i.e. that the header must state json?

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemReaderAutoConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+		return new KafkaItemReaderBuilder<Object, Map<Object, Object>>()
+				.partitions(kafkaItemReaderProperties.getPartitions())
+				.consumerProperties(consumerProperties)
+				.name(kafkaItemReaderProperties.getName())
+				.pollTimeout(Duration
+						.ofSeconds(kafkaItemReaderProperties.getPollTimeOutInSeconds()))
+				.saveState(true).topic(kafkaItemReaderProperties.getTopic()).build();
+	}
+
+	private void validateProperties(KafkaItemReaderProperties kafkaItemReaderProperties) {
```
benas, line 74: `Key.deserializer` and `value.deserializer` are mandatory properties to build a `KafkaItemReader` (see https://github.com/spring-projects/spring-batch/blob/15a393b8ef836aa642f9e1bb94c561095351fe8f/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemReader.java#L99-L102). So unless we are sure `kafkaProperties.getConsumer().buildProperties()` will contain them, we need to validate they are present here.

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemReaderAutoConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+		consumerProperties.putAll(this.kafkaProperties.getConsumer().buildProperties());
+		validateProperties(kafkaItemReaderProperties);
+		if (kafkaItemReaderProperties.getPartitions() == null
+				|| kafkaItemReaderProperties.getPartitions().size() == 0) {
+			kafkaItemReaderProperties.setPartitions(new ArrayList<>(1));
+			kafkaItemReaderProperties.getPartitions().add(0);
+		}
+		return new KafkaItemReaderBuilder<Object, Map<Object, Object>>()
+				.partitions(kafkaItemReaderProperties.getPartitions())
+				.consumerProperties(consumerProperties)
```
benas, line 67: I see `consumerProperties` are initialized from `this.kafkaProperties.getConsumer().buildProperties()` then passed to the reader. Please note that Spring Batch uses kafka constants like `ConsumerConfig.GROUP_ID_CONFIG`, `ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG`, etc for properties names which resolve respectively to `group.id`, `value.deserializer`, etc. Looking at the tests (and if I'm not mistaken), properties coming from `this.kafkaProperties.getConsumer().buildProperties()` will have a prefix `spring.kafka.consumer.` like `spring.kafka.consumer.group-id=1` and `"spring.kafka.consumer.value-deserializer="+ JsonDeserializer.class.getName()` . Those will not be accepted by the `KafkaItemReader`'s constructor or its builder.

cppwfs: Hello @benas ,
In this case the kafkaProperties.getConsumer.BuildProperties sets the properties with the appropriate prefix.  As shown here: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java#L420

Now another question.   If users are used to using the ConsumerProperties prefix instead of the Spring Kafka prefix(ala boot style properties), we may want to support those here as well (as overrides).   Thoughts?

benas: If these single-step job starters are intended for boot users primarily, I would use the boot style first.

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemReaderAutoConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.kafka.KafkaProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.util.StringUtils;
+
+/**
+ *
+ * Autconfiguration for a {@code KafkaItemReader}.
```
benas, line 39: typo here: should be `AutoConfiguration`

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemReaderAutoConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+			kafkaItemReaderProperties.setPartitions(new ArrayList<>(1));
+			kafkaItemReaderProperties.getPartitions().add(0);
+		}
+		return new KafkaItemReaderBuilder<Object, Map<Object, Object>>()
+				.partitions(kafkaItemReaderProperties.getPartitions())
+				.consumerProperties(consumerProperties)
+				.name(kafkaItemReaderProperties.getName())
+				.pollTimeout(Duration
+						.ofSeconds(kafkaItemReaderProperties.getPollTimeOutInSeconds()))
+				.saveState(true).topic(kafkaItemReaderProperties.getTopic()).build();
```
benas, line 71: `saveState` is hard-coded to `true` here. Is this intended? I think it should be configurable like other readers.

spring-cloud-starter-single-step-batch-job/src/test/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemReaderAutoConfigurationTests.java
```diff
@@ -0,0 +1,240 @@
...
+		producer.send(new ProducerRecord<>(topic, "my-aggregate-id", testMap));
+		producer.flush();
+		producer.close();
+	}
+
+	@EnableBatchProcessing
+	@Configuration
+	public static class CustomMappingConfiguration {
+		@Bean
+		public ListItemWriter<Map> itemWriter() {
```
benas, line 235: This could be:

```suggestion
		public ListItemWriter<Map<Object, Object>> itemWriter() {
```

to avoid the `Raw use of parameterized class 'Map'` warning.

cppwfs:
Updated based on code review.
Thanks @benas for all your help on this.
Rebased, Polished, Squashed, Merged

PR #689, opened 2020-07-22
Add AMQPItemReader

cppwfs:
resolves TASK-680A

When running an application that uses the AmqpItemReader single step job, be sure to set the `spring.cloud.task.closecontextEnabled` to `true`


spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/rabbit/AmqpItemReaderAutoConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+					"DefaultReceiveQueue must not be empty nor null");
+		}
+		return new Queue(this.rabbitProperties.getTemplate().getDefaultReceiveQueue(),
+				true);
+	}
+
+	@Bean
+	public AmqpItemReader<Map<Object, Object>> amqpItemReader(AmqpTemplate amqpTemplate) {
+		return new AmqpItemReaderBuilder<Map<Object, Object>>().amqpTemplate(amqpTemplate)
+				.build();
```
benas, line 74: The `AmqpItemReader` has an [itemType](https://docs.spring.io/spring-batch/docs/4.2.x/api/org/springframework/batch/item/amqp/AmqpItemReader.html#setItemType-java.lang.Class-) property, but it seems there is no way to configure it here. Is this intended?

spring-cloud-starter-single-step-batch-job/pom.xml
```diff
@@ -9,6 +9,12 @@
 
 	<artifactId>spring-cloud-starter-single-step-batch-job</artifactId>
 
+	<properties>
+		<test.containers.version>1.9.1</test.containers.version>
+		<test.rabbit.containers.version>1.12.5</test.rabbit.containers.version>
```
benas, line 14: There is a newer version of testcontainers (and its rabbitmq module): 1.14.3. Any reason not using the latest version?

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/rabbit/AmqpItemReaderAutoConfiguration.java
cppwfs: Good catch.   Added support and tests accordingly.

spring-cloud-starter-single-step-batch-job/pom.xml
cppwfs: Updated to latest version.

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/rabbit/AmqpItemReaderAutoConfiguration.java
```diff
@@ -0,0 +1,89 @@
...
+	private RabbitProperties rabbitProperties;
+
+	@Bean
+	public AmqpItemReaderProperties amqpItemReaderProperties() {
+		return new AmqpItemReaderProperties();
+	}
+
+	@ConditionalOnBean(RabbitProperties.class)
+	@Bean
+	public Queue myQueue() {
```
mminella, line 61: Rename this from `myQueue`.  That's a good name for a test...not for prod code ;)

mminella:
Rebased, squashed, and merged as 5fa4a2028de3119059b2b3cb3ec41945b192628b

PR #697, opened 2020-08-05
Adding KafkaItemWriter single step autoconfig

### LINKED ISSUE #696: Add KafkaItemWriter as Single Step Batch Job

cppwfs:
resolves #696
The base set of properties to setup are the following to activate the writer:
```
spring.batch.job.kafkaitemwriter.topic=<your topic name>
spring.kafka.consumer.bootstrap-servers=localhost:9092
```
If you  want to have a serializer for the value or key that is different than the boot provided default `StringSerializer`  you  can set the following properties to the serializer needed.  In the example below I use the JsonSerializer.
```
spring.kafka.producer.key-serializer=org.springframework.kafka.support.serializer.JsonSerializer
spring.kafka.producer.value-serializer=org.springframework.kafka.support.serializer.JsonSerializer
```

If a user needs another itemKeyMapper besides the default provided you can create a new bean with the name `beanItemNameMapper`. i.e.
```
	@Bean(name = "batchItemKeyMapper")
	public Converter<Object, Object> batchItemKeyMapper()
```

### DIFF
--- a/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterAutoConfiguration.java
+++ b/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterAutoConfiguration.java
@@ -57,18 +57,25 @@ public class KafkaItemWriterAutoConfiguration {
 	@ConditionalOnProperty(prefix = "spring.batch.job.kafkaitemwriter", name = "name")
 	public KafkaItemWriter<Object, Map<Object, Object>> kafkaItemWriter(
 			KafkaItemWriterProperties kafkaItemWriterProperties,
-			ProducerFactory<Object, Map<Object, Object>> producerFactory) {
+			ProducerFactory<Object, Map<Object, Object>> producerFactory,
+			Converter<?, ?> itemKeyMapper) {
 
 		validateProperties(kafkaItemWriterProperties);
 		KafkaTemplate template = new KafkaTemplate(producerFactory);
 		template.setDefaultTopic(kafkaItemWriterProperties.getTopic());
 		return new KafkaItemWriterBuilder<Object, Map<Object, Object>>().delete(false)
-				.kafkaTemplate(template).itemKeyMapper(new Converter() {
-					@Override
-					public Object convert(Object source) {
-						return "";
-					}
-				}).build();
+				.kafkaTemplate(template).itemKeyMapper(itemKeyMapper).build();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public Converter defaultItemKeyMapper() {
+		return new Converter() {
+			@Override
+			public Object convert(Object source) {
+				return "";
+			}
+		};
 	}
 
 	@Bean
--- a/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterAutoConfiguration.java
+++ b/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterAutoConfiguration.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2020-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.task.batch.autoconfigure.kafka;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.batch.item.kafka.KafkaItemWriter;
+import org.springframework.batch.item.kafka.builder.KafkaItemWriterBuilder;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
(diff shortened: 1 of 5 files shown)


cppwfs:
Autowire in ItemKeyMapper instead of `null`

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterProperties.java
```diff
@@ -0,0 +1,71 @@
...
+ * @since 2.3
+ */
+@ConfigurationProperties(prefix = "spring.batch.job.kafkaitemwriter")
+public class KafkaItemWriterProperties {
+
+	private String name;
+
+	private String topic;
+
+	private long pollTimeOutInSeconds = 30L;
```
benas, line 34: This is not used on the writer side.

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterProperties.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Properties to configure a {@code KafkaItemWriter}.
+ *
+ * @author Glenn Renfro
+ * @since 2.3
+ */
+@ConfigurationProperties(prefix = "spring.batch.job.kafkaitemwriter")
+public class KafkaItemWriterProperties {
+
+	private String name;
```
benas, line 30: A `kafkaItemWriter` does not have a `name` property (it is not an `ItemSteam`), so this is not needed.

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterProperties.java
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * Returns the name of the topic from which messages will be read.
+	 * @return the name of the topic.
+	 */
+	public String getTopic() {
+		return topic;
+	}
+
+	/**
+	 * The topic name from which the messages will be read.
```
benas, line 64: On the writer's side, this should be "The topic name to which messages will be written". Same thing for the getter's javadoc.

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterAutoConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+	@ConditionalOnProperty(prefix = "spring.batch.job.kafkaitemwriter", name = "name")
+	public KafkaItemWriter<Object, Map<Object, Object>> kafkaItemWriter(
+			KafkaItemWriterProperties kafkaItemWriterProperties,
+			ProducerFactory<Object, Map<Object, Object>> producerFactory,
+			Converter<?, ?> itemKeyMapper) {
+
+		validateProperties(kafkaItemWriterProperties);
+		KafkaTemplate template = new KafkaTemplate(producerFactory);
+		template.setDefaultTopic(kafkaItemWriterProperties.getTopic());
+		return new KafkaItemWriterBuilder<Object, Map<Object, Object>>().delete(false)
```
benas, line 66: Why is `delete` hard-coded to `false` here? I think it should be configurable in `KafkaItemWriterProperties`.

cppwfs: DOH!    I have a branch I haven't pushed up with that change.

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterProperties.java
cppwfs: EEk, busted on copy paste :-)

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterProperties.java
cppwfs: We need the `name` to disable and enable auto configuration.

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterProperties.java
cppwfs: EeekEEk, busted on copy paste :-)

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/kafka/KafkaItemWriterAutoConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+				.kafkaTemplate(template).itemKeyMapper(itemKeyMapper).build();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public Converter defaultItemKeyMapper() {
+		return new Converter() {
+			@Override
+			public Object convert(Object source) {
+				return "";
```
benas, line 76: This might be problematic if the key type is not a String. I would return the source object as a key without any modification. BTW, since the itemkeyMapper is mandatory, I would make the user provide it and fail fast if it's absent. Otherwise, this default mapper can be silently used with unintended side effects (in this case, it would be writing all items with an empty String as a key).

cppwfs:
I also need to qualify the use of the Converter bean.   Boot out of the box enables 3, so I need to qualify the one we want.  A update is on the way.

cppwfs:
@benas Thanks for your help!
Rebased, Polished, Squashed, Merged

PR #699, opened 2020-08-06
Establish RabbitMQ Item Writer Single Step Batch

### LINKED ISSUE #698: Add AmqpItemWriter to singlestepbatch job

cppwfs:
resolves #698

1. To test you must first setup the exchange.   
* Create Queue `foo` to which you want the results.
* Create Exchange `fooexchange` as type topic.
* Bind `fooexchange` to `foo` queue with # for the routing key.

2. Set the properties in your app to use the exchange.  
*  spring.batch.job.rabbitamqpitemwriter.name=testwriter 
*  spring.cloud.task.closecontextEnabled=true
*  spring.rabbitmq.template.exchange=fooexchange

3.  Under the new commit we no longer use name to intialize the writer.  instead it is 
* spring.batch.job.rabbitamqpitemwriter.enabled=true


spring-cloud-starter-single-step-batch-job/pom.xml
```diff
@@ -41,6 +47,40 @@
 			<artifactId>h2</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.amqp</groupId>
+			<artifactId>spring-amqp</artifactId>
```
mminella, line 52: Dependencies like this should be optional so we don't bring in the whole world...

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/rabbit/RabbitAmqpItemWriterAutoConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+@EnableConfigurationProperties(RabbitAmqpItemWriterProperties.class)
+@AutoConfigureAfter(BatchAutoConfiguration.class)
+public class RabbitAmqpItemWriterAutoConfiguration {
+
+	@Bean
+	@ConditionalOnProperty(prefix = "spring.batch.job.rabbitamqpitemwriter",
+			name = "name")
+	public AmqpItemWriter<Map<Object, Object>> amqpItemWriter(
+			RabbitTemplate rabbitTemplate) {
+		rabbitTemplate.setMessageConverter(new Jackson2JsonMessageConverter());
```
mminella, line 48: This should be replacable. AKA I should be allowed to bring my own and use the `Jackson2JsonMessageConverter` only if one hasn't been provided.

mminella:
Also, when I use this in an app...the app no longer shuts down automatically. Is there something with Rabbit that will inhibit that behavior?

mminella:
Rebased, squashed, and merged as b79657d3989c3cadf77c934abf30e7bfe2ca1d76

PR #723, opened 2020-09-22
Code format change and prop update to fix build

cppwfs:


cppwfs:
Not going to change code format at this time.

PR #727, opened 2020-09-25
Add ability for user to utilize external DBs for Multi-DB sample

### LINKED ISSUE #690: Unable to run multiple-datasources as-is on local Spring Cloud Data Flow installation
Hi,
I cloned the repo to my hard disk and opened the multiple-datasources example with IntelliJ Idea. 
Without altering the source code, I ran mvn package to produce a *.jar.

I then registered the jar as an app using the Spring shell with this command:
```
app register --type task --name multiple-datasources --uri file://root/scdf/multiple-datasources-2.3.0-SNAPSHOT.jar
```
I created the task definition as shown:
Screenshot 2020-07-13 at 21 44 25

And tried to run it. Unfortunately, SCDF does not record start or stop times for the task execution, and looking into the logs vía the shell, I found the following relevant messages:

```
2020-07-22 23:00:59.298 ERROR 140 --- [           main] o.s.c.t.listener.TaskLifecycleListener   : An event to end a task has been received for a task that has not yet started.
2020-07-22 23:00:59.298  WARN 140 --- [           main] s.c.a.AnnotationConfigApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.context.ApplicationContextException: Failed to start bean 'taskLifecycleListener'; nested exception is java.lang.IllegalArgumentException: Invalid TaskExecution, ID 52 not found
```

I've looked at similar issues, but am unable to understand what I'm doing wrong. Moreover I was hoping that the example would just run "out of the box". What am I missing?

Thanks!

cppwfs:
resolves #690

### DIFF
--- a/spring-cloud-task-samples/multiple-datasources/README.adoc
+++ b/spring-cloud-task-samples/multiple-datasources/README.adoc
@@ -11,7 +11,9 @@ which one to be used for the Spring Cloud Task repository.
 
 * `MultipleDataSourcesApplication` - the Spring Boot Main Application.
 * `SampleCommandLineRunner` - the `CommandLineRunner` implementation for this task.  It outputs the number of `DataSource` beans found in the context (should be 2).
-* `DataSourceConfiguration` - Configures two `DataSource` beans.
+* `EmbeddedDataSourceConfiguration` - Configures two `DataSource` beans using embedded databases.
+* `ExternalDataSourceConfiguration` - Configures two `DataSource` beans using external databases.
+
 * `CustomTaskConfigurer` - Uses a Spring `@Qualifier` to specify the correct `DataSource` to use.
 
 == Build:
@@ -21,9 +23,25 @@ which one to be used for the Spring Cloud Task repository.
 $ mvn clean package
 ----
 
-== Run:
+== Execute sample using 2 embedded databases (default):
 
 [source,shell,indent=2]
 ----
 $ java -jar target/multiple-datasources-2.2.2.RELEASE.jar
+----
+
+== Execute sample using 2 external databases:
+
+Using the `external` profile, users will be able to establish both the default `spring.datasource` data source and a `second.datasource` data source.
+For example:
+[source,shell,indent=2]
+----
+export spring_datasource_url=<your db url>
+export spring_datasource_username=<your db user name>
+export spring_datasource_password=<your db user password>
+export spring_datasource_driverClassName=org.mariadb.jdbc.Driver
+export second_datasource_url=jdbc:<your db url>
+export second_datasource_username=<your db user name>
+export second_datasource_password=<your db user password>
+java -jar target/multiple-datasources-2.3.0-RELEASE.jar --spring.profiles.active=external
 ----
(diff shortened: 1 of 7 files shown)


spring-cloud-task-samples/multiple-datasources/README.adoc
```diff
@@ -21,9 +23,25 @@ which one to be used for the Spring Cloud Task repository.
...
+For example:
+[source,shell,indent=2]
+----
+export spring_datasource_url=<your db url>
+export spring_datasource_username=<your db user name>
+export spring_datasource_password=<your db user password>
+export spring_datasource_driverClassName=org.mariadb.jdbc.Driver
+export second_datasource_url=jdbc:<your db url>
+export second_datasource_username=<your db user name>
+export second_datasource_password=<your db user password>
```
benas, line 45: I think `export second_datasource_driverClassName=<your second db driver class name>` is missing here. Do you agree?

spring-cloud-task-samples/multiple-datasources/src/main/resources/application.properties
```diff
@@ -1,2 +1,4 @@
 spring.application.name=Demo Multiple DataSources Task
 logging.level.org.springframework.cloud.task=DEBUG
+
+spring.profiles.default=embedded
```
benas, line 4: IntelliJ is showing a warning here: `Cannot resolve configuration property 'spring.profiles.default' `. Should this be `spring.profiles.active=embedded` as in https://github.com/spring-cloud/spring-cloud-task/pull/727/files#diff-18da3e6827339e29aef887fbb5946d77bb4b38966bd0f636c83b0b37937a3a2fR38 ?

spring-cloud-task-samples/multiple-datasources/README.adoc
```diff
@@ -21,9 +23,25 @@ which one to be used for the Spring Cloud Task repository.
 $ mvn clean package
 ----
 
-== Run:
+== Execute sample using 2 embedded databases (default):
 
 [source,shell,indent=2]
 ----
 $ java -jar target/multiple-datasources-2.2.2.RELEASE.jar
```
benas, line 30: This is not included in the PR, but the version of the jar in the command should be the same whether we use the `embedded` or `external` profile.

spring-cloud-task-samples/multiple-datasources/src/main/java/io/spring/configuration/ExternalDataSourceConfiguration.java
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2018-2019 the original author or authors.
```
benas, line 2: The year range should be `2018-2020` (same for other files).

spring-cloud-task-samples/multiple-datasources/README.adoc
cppwfs: Updated.   Thanks for catching

spring-cloud-task-samples/multiple-datasources/README.adoc
cppwfs: Ooops.   Thanks for catching

spring-cloud-task-samples/multiple-datasources/src/main/resources/application.properties
cppwfs: Good catch.  Thanks!

spring-cloud-task-samples/multiple-datasources/src/main/java/io/spring/configuration/ExternalDataSourceConfiguration.java
cppwfs: 2020 already? :-D   Thanks for catching

PR #738, opened 2020-10-29
Update versions for jacoco and rabbit versions for tests

cppwfs:
Update versions for testcontainer andd rabbit versions for tests
Update jacoco so that it will build properly with the latest maven version.


mminella:
LGTM. Merged

PR #740, opened 2020-11-03
Added sort validation

### LINKED ISSUE #739: Add sort validation to JdbcTaskExecutionDao
Currently the value passed into the `JdbcTaskExecutionDao` via a `Pageable` is not validated. This issue is to add validation to that parameter.

mminella:
This commit now validates that the value passed via a PageRequest to
sort the results by is a valid value.

Resolves #739

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/dao/JdbcTaskExecutionDao.java
```diff
@@ -161,6 +163,21 @@
 
 	private DataFieldMaxValueIncrementer taskIncrementer;
 
+	private static final Set<String> validSortColumns = new HashSet<>(10);
+
+	static {
+		validSortColumns.add("TASK_EXECUTION_ID");
```
cppwfs, line 169: Do we want these checks to be case insenstive.   i.e. lowercase as well as uppercase.

mminella: Good catch...updated.

cppwfs:
LGTM

cppwfs:
Rebased Squashed merged

PR #743, opened 2020-12-01
Updated generics on single step batch job starter

mminella:
This commit updates the generics for the domain object used in the
single step batch job starter from a 'Map<Object, Object>' to a
'Map<String, Object>'. This better reflects what is actually used.


cppwfs:
LGTM.   Let me kick the tires on some scenarios, if no issues pop up I'll merge.

PR #751, opened 2020-12-11
Replaced CommandLineRunner in examples with ApplicationRunner in docs

### LINKED ISSUE #749: Update Task Docs to replace CommandLineRunner with ApplicationRunner
We need to replace CommandLineRunner in our coding samples with ApplicationRunner.

cppwfs:
Also removed reference to CloudFoundry from docs.   This is covered in dataflow.spring.io.

resolves #749


cppwfs:
Leaving as is

PR #777, opened 2021-03-26
Removed TaskJobCommandLineRunner in prep for its removal from boot

cppwfs:
…boot


cppwfs:
Debrecating this PR for Issue #787

PR #783, opened 2021-05-07
Updating task execution id next val to use sequence for sqlserver

cppwfs:
resolves TASK-782

### DIFF
--- a/docs/src/main/asciidoc/appendix-task-repository-schema.adoc
+++ b/docs/src/main/asciidoc/appendix-task-repository-schema.adoc
@@ -83,3 +83,15 @@ Used for the `single-instance-enabled` feature discussed <<features-single-insta
 
 NOTE: The DDL for setting up tables for each database type can be found https://github.com/spring-cloud/spring-cloud-task/tree/master/spring-cloud-task-core/src/main/resources/org/springframework/cloud/task[here].
 --
+
+=== SQL Server
+By default Spring Cloud Task uses a sequence table for determining the `TASK_EXECUTION_ID` for the `TASK_EXECUTION` table.
+However, when launching multiple tasks simultaneously while using SQL Server, this can cause a deadlock to occur on the `TASK_SEQ` table.
+The resolution is to drop the `TASK_EXECUTION_SEQ` table and create a sequence using the same name.   For example:
+```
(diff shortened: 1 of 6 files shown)


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/repository/support/TaskSqlServerSequenceMaxValueIncrementer.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.sql.DataSource;
+
+import org.springframework.jdbc.support.incrementer.AbstractSequenceMaxValueIncrementer;
+
+/**
+ * Incrementer using SQL Server's sequence.
+ * @author Glenn Renfro
+ * @since 2.3.2
+ */
+public class TaskSqlServerSequenceMaxValueIncrementer extends AbstractSequenceMaxValueIncrementer {
```
mminella, line 28: See previous note about removing the `Task` prefix from the name of this class. Also...unit test for this class?

mminella (changes requested):

docs/src/main/asciidoc/appendix-task-repository-schema.adoc
```diff
@@ -83,3 +83,15 @@ Used for the `single-instance-enabled` feature discussed <<features-single-insta
 
 NOTE: The DDL for setting up tables for each database type can be found https://github.com/spring-cloud/spring-cloud-task/tree/master/spring-cloud-task-core/src/main/resources/org/springframework/cloud/task[here].
 --
+
+=== SQL Server
+By default Spring Cloud Task uses a sequence table for determining the `TASK_EXECUTION_ID` for the `TASK_EXECUTION` table.
+However, when launching multiple tasks simultaneously while using SQL Server, this can cause a deadlock to occur on the `TASK_SEQ` table.
+The resolution is to drop the `TASK_EXECUTION_SEQ` table and create a sequence using the same name.   For example:
```
mminella, line 90: Do we want to call out that existing users should set the start with value to be one more than the current value in the sequence table?

cppwfs:
Rebased, Squashed and Merged with permission of @mminella

PR #784, opened 2021-05-17
Mark TaskRepositoryInitializer as not lazy

hpoettker:
Spring Cloud Task apps currently do not work as expected when Spring Boot lazy initialization is switched on. When the database is not initialized, they fail with the following error even if database initialization is explicitly enabled:

```
org.springframework.context.ApplicationContextException: Failed to start bean 'taskLifecycleListener';
nested exception is org.springframework.dao.DataAccessResourceFailureException: Could not obtain sequence value;
nested exception is org.h2.jdbc.JdbcSQLSyntaxErrorException: Syntax error in SQL statement "SELECT TASK_SEQ.NEXTVAL FROM[*] DUAL";
expected "identifier";
SQL statement: select TASK_SEQ.nextval from dual [42001-200]
```

The problem can be reproduced with any SCT app, e.g. with the minimal example from the github readme page, when `spring.main.lazy-initialization=true` is set in `application.properties`.

The issue is more relevant for integration testing and test environments and can be worked around by users with a `LazyInitializationExcludeFilter`. But it would be nice to have consistent behavior independent of whether initialization is lazy or not.

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskAutoConfiguration.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/SimpleTaskAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 the original author or authors.
+ * Copyright 2015-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


cppwfs:
LGTM, Thank you for your contribution!

cppwfs:
Thank you for your contribution!

Issue #793, opened 2021-08-06
partitioned-batch-job raises NPE when used with mutiple Partiton handlers


nicolasduminil:
Hello,

I modified the partitioned-batch-job sample such that to add the following method:

      @Bean
      public PartitionHandler partitionHandler2(TaskLauncher taskLauncher, JobExplorer jobExplorer, TaskRepository taskRepository)
      {
        Resource resource = this.resourceLoader.getResource("maven://fr.simplex_software.tests:partitioned-job:1.0-SNAPSHOT");
        DeployerPartitionHandler partitionHandler = new DeployerPartitionHandler(taskLauncher, jobExplorer, resource, "workerStep");
        List<String> commandLineArgs = new ArrayList<>(3);
        commandLineArgs.add("--spring.profiles.active=worker");
        commandLineArgs.add("--spring.cloud.task.initialize-enabled=false");
        commandLineArgs.add("--spring.batch.initializer.enabled=false");
        partitionHandler.setCommandLineArgsProvider(new PassThroughCommandLineArgsProvider(commandLineArgs));
        partitionHandler.setEnvironmentVariablesProvider(new SimpleEnvironmentVariablesProvider(this.environment));
        partitionHandler.setMaxWorkers(2);
        partitionHandler.setApplicationName("PartitionedBatchJobTask");
        return partitionHandler;
      }

Running the job raises the following exception:

    2021-08-06 13:12:47.071 DEBUG 13084 --- [           main] o.s.c.t.r.support.SimpleTaskRepository   : Creating: 
    TaskExecution{executionId=149, parentExecutionId=null, exitCode=null, taskName='null', startTime=null, endTime=null, 
    exitMessage='null', externalExecutionId='null', errorMessage='null', arguments=[]}
    2021-08-06 13:12:47.074 ERROR 13084 --- [           main] o.s.batch.core.step.AbstractStep         : Encountered an error 
    executing step step1 in job partitionedJob-494091066

    java.lang.NullPointerException: null
         at 
        org.springframework.cloud.task.batch.partition.DeployerPartitionHandler.launchWorker(DeployerPartitionHandler.java:347) 
        ~[spring-cloud-task-batch-2.3.3.jar:2.3.3]
         at 
        org.springframework.cloud.task.batch.partition.DeployerPartitionHandler.launchWorkers(DeployerPartitionHandler.java:313) 
        ~ [spring-cloud-task-batch-2.3.3.jar:2.3.3]
        ...

[...]

PR #796, opened 2021-08-15
Support multiple task listener beans of same type

isikerhan:
Hello,

In a Spring Cloud Task application, when I have more than one beans of task listeners with @BeforeTask, @AfterTask and/or @FailedTask of same type (or different types inheriting the same @BeforeTask/@AfterTask/@FailedTask method), only single one of those listener beans work. Listener methods of the rest of the beans are never invoked. Please see the following example.

```java
public class TaskListener {

    @BeforeTask
    public void beforeTask(TaskExecution taskExecution) {
        // ...
    }
}

@Configuration
public class ListenerConfiguration {
    
    // this bean's beforeTask method is never called
    @Bean
    public TaskListener taskListener() {
        return new TaskListener();
    }

    @Bean
    public TaskListener otherTaskListener() {
        return new TaskListener();
    }
}

```
I don't know if this is the intended behaviour but this is an unexpected behaviour to me and not documented AFAIK. This behaviour also causes multi-step partitioned tasks to fail with NPE since DeployerPartitionHandler uses @BeforeTask to inject TaskExecution. (See #793)

This PR changes this behaviour so that all listener methods are invoked even the listener beans have the same type (or different types inheriting the same listener method) I've added some unit tests as well to cover "multiple listener beans of same type" case. I've also tested these changes with a multi-step partitioned task and it just seemed to work fine.

Thanks.

### DIFF
--- a/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskListenerExecutorObjectFactory.java
+++ b/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskListenerExecutorObjectFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 the original author or authors.
+ * Copyright 2018-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


cppwfs:
@isikerhan Thank you for this contribution!   If you can sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring) I can merge this PR.

cppwfs:
Ahh.   You have already signed it.  Sorry I missed that!

PR #798, opened 2021-09-07
Resolves Circular Dependency issue

### LINKED ISSUE #797: Circular reference error when autowiring JobBuilderFactory in task applications
As discussed in https://github.com/spring-projects/spring-batch/issues/3991 if a user includes a bean required for a `JobBuilderFactory` a BeanCurrentlyInCreationException.   This also affects `JdbcCursorItemReaderAutoConfiguration` and `FlatFileItemReaderAutoConfiguration`.

cppwfs:
Makes sure that dependencies are passed in via parameters in bean declaraction vs. Autowiring

refer to https://github.com/spring-projects/spring-batch/issues/3991
resolves #797

### DIFF
--- a/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/flatfile/FlatFileItemReaderAutoConfiguration.java
+++ b/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/flatfile/FlatFileItemReaderAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2020 the original author or authors.
+ * Copyright 2019-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -41,6 +41,7 @@ import org.springframework.context.annotation.Configuration;
  * Autconfiguration for a {@code FlatFileItemReader}.
  *
  * @author Michael Minella
+ * @author Glenn Renfro
  * @since 2.3
  */
 @Configuration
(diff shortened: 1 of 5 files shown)


spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/flatfile/FlatFileItemReaderAutoConfiguration.java
```diff
@@ -84,24 +75,24 @@ public FlatFileItemReaderAutoConfiguration(FlatFileItemReaderProperties properti
 				.comments(this.properties.getComments()
 						.toArray(new String[this.properties.getComments().size()]));
 
-		if (this.lineTokenizer != null) {
-			mapFlatFileItemReaderBuilder.lineTokenizer(this.lineTokenizer);
+		if (lineTokenizer != null) {
+			mapFlatFileItemReaderBuilder.lineTokenizer(lineTokenizer);
```
mminella, line 79: Why add the if statements? We previously depended upon the builder to validate that the state was valid.

cppwfs: They were not required so , they can be null.   So we need the null check.

mminella: Again...the builder does that check. We don't do null checks on any of the other places we do this kind of thing (see [here](https://github.com/spring-cloud/spring-cloud-task/pull/798/files#diff-af8f3bbf8f60a5583ee99cc0d35b7938d5838ccdcb4f8361a979fb4e16e96748R75) for example).

cppwfs: Bye bye `if`s

PR #800, opened 2021-09-16
Spring Cloud Task Stream is not supported by singlestep batch

### LINKED ISSUE #799: SingleStepBatchJob with spring-cloud-task-stream enables Kafka client
If both `spring-cloud-starter-single-step-batch-job` and `spring-cloud-task-stream` are added as dependencies to a spring cloud task application the Kafka Client from Spring Boot is enabled even if neither the reader nor writer are for Kafka.   
Need to disable BatchEventAutoConfiguration and TaskEventAutoConfiguration when using SingleStepBatchJob.

cppwfs:
resolves #799


cppwfs:
Rebased Squashed Merged

PR #832, opened 2022-02-16
Add default methods in `TaskExecutionListener`

hpoettker:
The PR adds empty default implementations to `TaskExecutionListener` and deprecates the class `TaskExecutionListenerSupport` that currently can be extended to avoid implementing empty methods in implementations.

An analogous change has been done in Spring Batch: spring-projects/spring-batch/issues/3924


cppwfs:
Rebased Merged.
Again thanks for the contribution!

PR #834, opened 2022-02-28
Support asynchronous launch of partitions.

### LINKED ISSUE #785: Slow Batch Partitioner task launches with worker count
Currently Launching of batch partition workers is done sequentially as shown here: https://github.com/spring-cloud/spring-cloud-task/blob/7f20c5a06ed7b1edb83b8bd187fcdaccfe8ed20e/spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java#L309
Meaning the code blocks until it gets a successful launch from the deployer.   This can slow the deployment of partitions by multiple minutes.  Explore launching partitions in parallel(async) not sequentially (as currently implemented).   In this case we can use replace the `for` loop to a set of tasks submitted in parallel to a task executor. 
And then if of the task launches fails we can throw an exception that can be handled or floated to the top  and be handled by the manager .
This issue was identified here: https://stackoverflow.com/questions/68117167/spring-cloud-data-flow-asynchronous-deploymentpartitionhanlder

cppwfs:
Resolves #785

Updated to allow user to set ThreadPoolTaskExecutor.


spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
```diff
@@ -145,19 +141,23 @@
 
 	private boolean defaultArgsAsEnvironmentVars = false;
 
+	private ThreadPoolTaskExecutor threadPoolTaskExecutor;
```
benas, line 144: I would program to an interface here (like `TaskExecutor`) and not a specific implementation. The user can then choose which implementation to use to launch tasks. Do you agree?

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
```diff
@@ -145,19 +141,23 @@
...
+	 * @param resource The {@link Resource} to the app to be launched.
+	 * @param stepName The name of the step.
+	 * @param threadPoolTaskExecutor If task launches should occur asynchronously then provide a {@link ThreadPoolTaskExecutor}.  Default is null.
 	 */
-
-	public DeployerPartitionHandler(TaskLauncher taskLauncher, JobExplorer jobExplorer,
-			Resource resource, String stepName, TaskRepository taskRepository) {
+	public DeployerPartitionHandler(org.springframework.cloud.deployer.spi.task.TaskLauncher taskLauncher, JobExplorer jobExplorer,
+		Resource resource, String stepName, TaskRepository taskRepository,
+		ThreadPoolTaskExecutor threadPoolTaskExecutor) {
```
benas, line 160: Same here, the API should work against an interface.

spring-cloud-task-batch/src/main/java/org/springframework/cloud/task/batch/partition/DeployerPartitionHandler.java
```diff
@@ -169,6 +169,19 @@ public DeployerPartitionHandler(TaskLauncher taskLauncher, JobExplorer jobExplor
...
+	/**
+	 * Constructor initializing the DeployerPartitionHandler instance.
+	 * @param taskLauncher The {@link org.springframework.cloud.deployer.spi.task.TaskLauncher} used to execute partitioned tasks.
+	 * @param jobExplorer The {@link JobExplorer} to acquire the status of the job.
+	 * @param resource The {@link Resource} to the app to be launched.
+	 * @param stepName The name of the step.
+	 */
+	public DeployerPartitionHandler(org.springframework.cloud.deployer.spi.task.TaskLauncher taskLauncher, JobExplorer jobExplorer,
+		Resource resource, String stepName, TaskRepository taskRepository) {
+		this(taskLauncher, jobExplorer, resource, stepName, taskRepository, null);
```
benas, line 184: The default `TaskExecutor` could be a `SyncTaskExecutor` to preserve the current default behaviour of launching tasks sequentially.

cppwfs:
Squashed Merged Polished

Issue #138, opened 2016-05-06
Reconsider to use @TestPropertySource for common integration test properties


sabbyanandan:
_From @artembilan on April 1, 2016 15:19_

See https://github.com/spring-projects/spring-boot/issues/5555 for more information

_Copied from original issue: spring-cloud/spring-cloud-stream-modules#247_

sabbyanandan:
This issue was moved to spring-cloud/spring-cloud-stream-app-starters#86

sabbyanandan:
Moved to the wrong project! 😵

PR #836, opened 2022-03-07
Updating task metrics to micrometer 2.0

cppwfs:
Changes include:
* Use Observable instead of Timer
* Update the setting of lowCardinalityTag to the correct events.
* Update the docs
* Remove EXTERNAL_EXECUTION_ID
* Currently using longTaskTimer until Observable supports this feature (or doesn't).
* Utilize registry provided by boot instead of globalregistry
* Set the handler until Boot does this as shown on line 80 of TaskLifecycleConfiguration.


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -146,7 +149,9 @@ public TaskLifecycleListener(TaskRepository taskRepository,
 		this.taskExplorer = taskExplorer;
 		this.taskProperties = taskProperties;
 		this.taskListenerExecutorObjectFactory = taskListenerExecutorObjectFactory;
-		this.taskMetrics = new TaskMetrics();
+		taskMetrics = meterRegistry != null ?
+			new TaskMetrics(meterRegistry) : null;
+		this.taskMetrics = taskMetrics;
```
jonatan-ivanov, line 154: Wouldn't this be the same as:
```
this.taskMetrics = meterRegistry != null ? new TaskMetrics(meterRegistry) : null;
```
so that we can also remove `TaskMetrics taskMetrics;` from line\#138

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskMetrics.java
```diff
@@ -88,41 +92,47 @@
...
-	private Throwable exception;
+	private Observation observation;
 
 	public void onTaskStartup(TaskExecution taskExecution) {
-		LongTaskTimer longTaskTimer = LongTaskTimer
+		if (this.registry != null) {
+			this.observation = Observation.start(SPRING_CLOUD_TASK_METER, this.registry)
+				.lowCardinalityTag(TASK_EXIT_CODE_TAG, String.valueOf(taskExecution.getExitCode()))
+				.lowCardinalityTag(TASK_STATUS_TAG, STATUS_SUCCESS)
+				.lowCardinalityTag(TASK_EXCEPTION_TAG, "none")
```
jonatan-ivanov, line 104: There is an `observation.error(Throwable)` method, which does the same (and more), can we utilize that instead?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskMetrics.java
```diff
@@ -88,41 +92,47 @@
 	 */
 	public static final String TASK_EXCEPTION_TAG = "task.exception";
 
-	private Timer.Sample taskSample;
-
 	private LongTaskTimer.Sample longTaskSample;
 
-	private Throwable exception;
+	private Observation observation;
```
jonatan-ivanov, line 97: In order to make context propagation work, we should create a `Scope`.
Instead of storing the `Observation` here, I think we can open a `Scope` (`observation.openScope()`), store the scope here and close it in `onTaskEnd` (we can also get the observation from the scope: `scope.getCurrentObservation()`).

jonatan-ivanov (commented):
Looks good!

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
cppwfs: Last time I listen to the recommendation from Intellij :-D

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskMetrics.java
cppwfs: Cool, I'll give it a try.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskMetrics.java
```diff
@@ -88,41 +92,47 @@
...
 
 	public void onTaskStartup(TaskExecution taskExecution) {
-		LongTaskTimer longTaskTimer = LongTaskTimer
+		if (this.registry != null) {
+			this.observation = Observation.start(SPRING_CLOUD_TASK_METER, this.registry)
+				.lowCardinalityTag(TASK_EXIT_CODE_TAG, String.valueOf(taskExecution.getExitCode()))
```
marcingrzejszczak, line 102: We were supposed to migrate this to using `TagsProvider` approach - check https://github.com/spring-projects/spring-batch/pull/4065/files for reference

cppwfs: @marcingrzejszczak @jonatan-ivanov I updated the PR based on your code review comments.   First off thanks for the review!    

One note:  Since task does not have any High Cardinality tags and one is required, a `TagKey` for  HighCardinality tags was added.  It only contains a task.name tag.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/micrometer/TaskMetricsTests.java
```diff
@@ -43,15 +45,15 @@
...
-		taskMetrics = new TaskMetrics();
+		this.simpleMeterRegistry = (SimpleMeterRegistry) new SimpleMeterRegistry().withTimerObservationHandler();
+		this.taskMetrics = new TaskMetrics(this.simpleMeterRegistry);
 	}
 
 	@AfterEach
 	public void after() {
-		Metrics.globalRegistry.getMeters().forEach(Metrics.globalRegistry::remove);
+		this.simpleMeterRegistry.clear();
+		this.simpleMeterRegistry.setCurrentObservation(null);
```
marcingrzejszczak, line 55: Actually shouldn't the instrumentation handle setting the current observation to `null` ? I mean if you've done both `onTaskStartup` and `onTaskEnd` then there MUST not be any current observation. So instead of setting this to null you should do an assertion that current observation is null.

spring-cloud-task-core/src/test/java/org/springframework/cloud/task/micrometer/TaskMetricsTests.java
```diff
@@ -43,15 +45,15 @@
...
+		ObservationRegistryAssert.assertThat(this.simpleMeterRegistry).doesNotHaveRemainingObservation();
```
marcingrzejszczak, line 56: Actually since you've set this to null one line above this will be always true. You should remove line no 55 and leave this as a check that no observations got leaked

cppwfs:
@marcingrzejszczak @jonatan-ivanov I think we are getting close.  I updated the PR with the latest requested changes,

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskMetrics.java
```diff
@@ -125,11 +125,11 @@ public void onTaskFailed(Throwable throwable) {
...
-			if (this.scope != null) {
-				this.taskObservationContext.getTaskExecution().setExitCode(taskExecution.getExitCode());
-				this.scope.getCurrentObservation().stop();
-				this.scope = null;
-			}
+			this.taskObservationContext.getTaskExecution().setExitCode(taskExecution.getExitCode());
+			this.scope.getCurrentObservation().stop();
+			this.scope = null;
+			this.registry.getCurrentObservation().stop();
+			this.registry.setCurrentObservation(null);
```
marcingrzejszczak: Why are you nulling this in this line? Line 129 should be doing this with the `this.scope.close()` method. In general when `onTaskStart` creates an observation then `onTaskEnd` should stop it and remove it from thread local via `this.scope.close()`

jonatan-ivanov (approved):

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/TaskLifecycleConfiguration.java
```diff
@@ -62,17 +62,27 @@
...
 	@Autowired
 	public TaskLifecycleConfiguration(TaskProperties taskProperties,
-			ConfigurableApplicationContext context, TaskRepository taskRepository,
-			TaskExplorer taskExplorer, TaskNameResolver taskNameResolver,
-			ObjectProvider<ApplicationArguments> applicationArguments) {
+		ConfigurableApplicationContext context, TaskRepository taskRepository,
+		TaskExplorer taskExplorer, TaskNameResolver taskNameResolver,
+		ObjectProvider<ApplicationArguments> applicationArguments,
+		@Autowired(required = false) io.micrometer.core.instrument.MeterRegistry meterRegistry,
+		@Autowired(required = false) ObservationRegistry observationRegistry) {
```
marcingrzejszczak, line 75: You can do a `ObservationRegistry.NOOP` instead of setting `null`. That way you won't need any `null` checks.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java
```diff
@@ -132,7 +134,9 @@
 	public TaskLifecycleListener(TaskRepository taskRepository,
 			TaskNameResolver taskNameResolver, ApplicationArguments applicationArguments,
 			TaskExplorer taskExplorer, TaskProperties taskProperties,
-			TaskListenerExecutorObjectFactory taskListenerExecutorObjectFactory) {
+			TaskListenerExecutorObjectFactory taskListenerExecutorObjectFactory,
+			@Autowired(required = false) MeterRegistry meterRegistry, @Autowired(required = false) ObservationRegistry observationRegistry) {
```
marcingrzejszczak, line 138: you can set `ObservationRegistry.NOOP` instead of `null`

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskMetrics.java
```diff
@@ -55,74 +63,80 @@
...
+
+	private TaskExecutionTagsProvider tagsProvider = new DefaultTaskExecutionTagsProvider();
+
+	private TaskExecutionObservationContext taskObservationContext;
 
 	public void onTaskStartup(TaskExecution taskExecution) {
-		LongTaskTimer longTaskTimer = LongTaskTimer
+		if (this.registry != null) {
+			this.taskObservationContext = new TaskExecutionObservationContext(taskExecution);
+			Observation observation = Observation.start(SPRING_CLOUD_TASK_METER, this.taskObservationContext, this.observationRegistry)
```
marcingrzejszczak, line 110: You can have `ObservationRegistry` without `meterRegistry` theoretically. Just saying ;)

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskMetrics.java
```diff
@@ -55,74 +63,80 @@
...
-		}
-
-		if (this.longTaskSample != null) {
-			this.longTaskSample.stop();
-			this.longTaskSample = null;
+		if (this.scope != null) {
+			this.taskObservationContext.getTaskExecution().setExitCode(taskExecution.getExitCode());
+			this.scope.getCurrentObservation().stop();
+			this.observationRegistry.getCurrentObservation().stop();
+			this.observationRegistry.setCurrentObservation(null);
```
marcingrzejszczak, line 134: Instead of 

```java
this.scope.getCurrentObservation().stop();
			this.observationRegistry.getCurrentObservation().stop();
			this.observationRegistry.setCurrentObservation(null);
```

you can do 

```java
this.scope.getCurrentObservation().stop();
this.scope.close();
```

docs/src/main/asciidoc/features.adoc
```diff
@@ -450,3 +450,27 @@ In these cases the context is held open because a thread has been allocated
 set the `spring.cloud.task.closecontextEnabled` property to `true` when launching your task.
 This will close the application's context once the task is complete.
 Thus allowing the application to terminate.
+
+[[enable-task-metrics]]
+=== Enable Task Metrics
+Spring Cloud Task integrates with Micrometer and collects metrics for both `Observation` and `LongTaskTimer`.
+To enable Task metrics integration, you must add the `spring-boot-starter-actuator`,
+preferred registry implementation, and the micrometer core to your task application.
+An example maven set of dependencies to enable task metrics using Influx would be:
```
jonatan-ivanov, line 459: How about calling out that this is not only about metrics but more?

```suggestion
[[enable-task-observability]]
=== Enable Task Observability
Spring Cloud Task integrates with Micrometer and creates observations for the Tasks it executes.
To enable Task Observability integration, you must add `spring-boot-starter-actuator`, your preferred registry implementation (if you want to publish metrics), and micrometer-tracing (if you want to publish tracing data) to your task application.
An example maven set of dependencies to enable task observability and metrics using Influx would be:
```

docs/src/main/asciidoc/features.adoc
```diff
@@ -450,3 +450,27 @@ In these cases the context is held open because a thread has been allocated
...
+    <groupId>org.springframework.boot</groupId>
+    <artifactId>spring-boot-starter-actuator</artifactId>
+</dependency>
+<dependency>
+    <groupId>io.micrometer</groupId>
+    <artifactId>micrometer-registry-influx</artifactId>
+    <version>1.10.0-SNAPSHOT</version>
+    <scope>runtime</scope>
+</dependency>
+<dependency>
```
jonatan-ivanov, line 472: Actuator should bring this in, I think we can remove it from the example (eventually?).

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/MetricsAutoConfiguration.java
```diff
@@ -1,136 +0,0 @@
...
-							.commonTags("cf.app.id", applicationId)
-							.commonTags("cf.app.name", applicationName)
-							.commonTags("cf.app.version", applicationVersion)
-							.commonTags("cf.instance.index", instanceIndex);
-				}
-
-				registry.config().commonTags("task.name", taskName)
-						.commonTags("task.execution.id", taskExecutionId)
-						.commonTags("task.external.execution.id", taskExternalExecutionId)
-						.commonTags("task.parent.execution.id", taskParentExecutionId);
```
jonatan-ivanov, line 123: Shouldn't we register a `ObservationFilter` to add these to the desired Observations like it was done for Meters here?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/DefaultTaskExecutionKeyValuesProvider.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author Glenn Renfro
+ * @since 3.0.0
+ */
+public class DefaultTaskExecutionKeyValuesProvider implements TaskExecutionKeyValuesProvider {
+
+	@Override
+	public KeyValues getLowCardinalityKeyValues(TaskExecutionObservationContext context) {
+		TaskExecution execution = context.getTaskExecution();
+		return KeyValues.of(TaskExecutionMeters.TaskTags.TASK_NAME_TAG.of(execution.getTaskName()),
```
jonatan-ivanov, line 35: Weren't these moved to the place when the `Observation` is created? If so, we might not need this class anymore.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskExecutionMeters.java
```diff
@@ -0,0 +1,116 @@
...
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.docs.DocumentedMeter;
+
+/**
+ * Enumeration for task execution metrics.
+ *
+ * @author Glenn Renfro
+ * @since 3.0.0
+ */
+public enum TaskExecutionMeters implements DocumentedMeter {
```
jonatan-ivanov, line 29: Should not this be a `DocumentedObservation`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskExecutionMeters.java
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+
+		@Override
+		public String getPrefix() {
+			return "spring.cloud.task";
+		}
+	};
+
+	public enum TaskTags implements KeyName {
```
jonatan-ivanov, line 51: How about `TaskKeyNames`?
```suggestion
	public enum TaskKeyNames implements KeyName {
```

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskExecutionMeters.java
```diff
@@ -0,0 +1,116 @@
...
+
+		/**
+		 * Task name measurement tag.
+		 */
+		TASK_NAME_TAG {
```
jonatan-ivanov, line 56: Since there are no tags on the Observations anymore, I would just remove those, e.g.: `TASK_NAME` instead of `TASK_NAME`.

cppwfs: Since we are not using TaskExecutionMeters, this class will be removed.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/MetricsAutoConfiguration.java
cppwfs: In Task 3.x we will not support CF tags.  They will be mantained in the 2.x line.

docs/src/main/asciidoc/features.adoc
cppwfs: Agreed

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskExecutionMeters.java
cppwfs: Removed the TaskExecutionMeters since it was not being used.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskObservations.java
```diff
@@ -0,0 +1,143 @@
...
+	private ObservationRegistry observationRegistry;
+
+	public TaskObservations(ObservationRegistry observationRegistry, TaskObservationCloudKeyValues taskObservationCloudKeyValues) {
+		this.observationRegistry = observationRegistry;
+		this.taskObservationCloudKeyValues = taskObservationCloudKeyValues;
+	}
+
+	private Observation.Scope scope;
+
+	private TaskExecutionKeyValuesProvider tagsProvider = new DefaultTaskExecutionKeyValuesProvider();
```
jonatan-ivanov, line 94: `keyValuesProvider`?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/DefaultTaskExecutionKeyValuesProvider.java
```diff
@@ -38,17 +36,20 @@ public class DefaultTaskExecutionKeyValuesProvider implements TaskExecutionKeyVa
...
 
 	@Override
 	public KeyValues getHighCardinalityKeyValues(TaskExecutionObservationContext context) {
+		return getKeyValuesForTaskExecution(context);
+	}
+
+	private KeyValues getKeyValuesForTaskExecution(TaskExecutionObservationContext context) {
 		TaskExecution execution = context.getTaskExecution();
-		return KeyValues.of(TASK_NAME, execution.getTaskName());
+		return KeyValues.of(
```
jonatan-ivanov, line 49: A keyvalue should have low or high cardinality, if all of these are low, the `getHighCardinalityKeyValues` method can be revoved.

PR #844, opened 2022-05-11
First pass on observation

marcingrzejszczak:


spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/observation/TaskDocumentedObservation.java
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		@Override
+		public KeyName[] getLowCardinalityKeyNames() {
+			return TaskRunnerTags.values();
+		}
+
+		@Override
+		public String getPrefix() {
+			return "spring.cloud.task";
```
cppwfs, line 40: This update creates a breaking change where the original metric names are no longer valid.   For example the original query I played with looked something like:
`SELECT "upper" FROM "spring_cloud_task"  GROUP BY "spring_cloud_task_name"`
Now it looks like:
`SELECT "upper" FROM "spring_cloud_task_execution" GROUP BY "task_name"`
Do we want to  transform the metrics (enabled by property) using https://micrometer.io/docs/concepts#_transforming_metrics as an option so that their original queries will still work?   What are your thoughts on this?

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/observation/TaskObservationContext.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2013-2021 the original author or authors.
```
cppwfs, line 2: Update to 2022.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/observation/TaskDocumentedObservation.java
marcingrzejszczak: That's very good point...

Actually we could go the other way round. Leave what we have and by default migrate to legacy. That's because that way tracing tags will be correct and we would by default set metrics back to the legacy setup (this would be an opt-out feature, meaning you would have to explicitly state that you want to use the new tagging for metrics).

WDYT?

cppwfs: That sounds great!

marcingrzejszczak: Actually... The observation metrics are created as an addition to the ones coming from `TaskMetrics`. This PR doesn't remove any "legacy" metrics, so you should have both. Are you observing something different?

cppwfs: Ok, that makes sense.    I will resolve the comment above then.

cppwfs:
Now I need to add documentation on these features.

spring-cloud-task-core/src/main/java/org/springframework/cloud/task/configuration/observation/ObservationApplicationRunnerBeanPostProcessor.java
```diff
@@ -0,0 +1,45 @@
...
+
+	private final BeanFactory beanFactory;
+
+	ObservationApplicationRunnerBeanPostProcessor(BeanFactory beanFactory) {
+		this.beanFactory = beanFactory;
+	}
+
+	@Override
+	public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
+		if (bean instanceof ApplicationRunner applicationRunner && !(bean instanceof ObservationApplicationRunner)) {
```
marcingrzejszczak, line 39: Come to think of it - if we're doing this wrapping, then if a user has their own type of an `ApplicationRunner` or the `CommandLineRunner` then we're changing their type and they won't be able to inject it. What we could do is to actually do the wrapping wherever the code is being called at runtime. WDYT?

cppwfs: True.    
Like here: https://github.com/spring-cloud/spring-cloud-task/blob/main/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java#L318 and
https://github.com/spring-cloud/spring-cloud-task/blob/main/spring-cloud-task-core/src/main/java/org/springframework/cloud/task/listener/TaskLifecycleListener.java#L341 ?

cppwfs:
LGTM!    Thanks @marcingrzejszczak for the contribution!

PR #845, opened 2022-05-25
Replace getContainerIpAddress by getHost

eddumelendez:
`getContainerIpAddress` is deprecated.


cppwfs:
Thanks again for the contribution!

PR #847, opened 2022-06-14
Observability metadata docs conditionally available

marcingrzejszczak:
we will enable the metadata documentation only if we're NOT in the release train documentation


cppwfs:
LGTM!   Thanks!

Issue #803, opened 2021-09-27
Establish independent data connection props for JDBC Single Step Batch Job


cppwfs:
Issue:
Currently if a user wants to select a JdbcCursorReader and a JdbcWriter  on the same single step batch job they must use the same data source as the task application.
A user should be able to select a  different Datasource for either the JdbcCursorReader and JdbcWriter

cppwfs:
Squashed Merged

PR #848, opened 2022-06-17
Users can specify different datasources for the jdbc-item reader and or jdbc-item writer

cppwfs:
Users can specify a unique data source for the JdbcCursorItemReader and or writer instead of using the default datasource.
If a jdbc-item-reader or jdbc-item-writer data source is not specified the default will be used by the reader or writer.

Updated docs

Updated readme docs

### DIFF
--- a/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/jdbc/JDBCSingleStepDataSourceAutoConfiguration.java
+++ b/spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/jdbc/JDBCSingleStepDataSourceAutoConfiguration.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2022-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.task.batch.autoconfigure.jdbc;
+
+import javax.sql.DataSource;
+
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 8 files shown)


cppwfs:
Hello @benas!
This is an excellent point. 
_**Background**_
Let's revisit the intent of the Single-step starter and the Issue.   It is to allow a user to configure a basic single-step application via properties and write no code for the basic implementation (unless they want to provide a processor).  An example would be to create a task definition using Spring Cloud Data Flow's single-step application it provides (much like the one in Spring Cloud Task Samples).  Then select from a basic set of readers and writers from properties and configure them to what needs to be done.    
However, if a user wants to go beyond the basics, they can create their own project which includes the SingleStepBatchJobStarter dependency, and as you mentioned above, choose a different reader or writer by creating their own beans accordingly.    

_**Revisit the issue**_
The goal of issue #803 is to enhance the JDBC reader and writer provided by SingleStepBatchJobStarter with the capability to read from and write to a different database(s) than the default database.

_**The question**_

1. Is this feature adding enhanced value to the JDBC ->JDBC case? Or is this new feature exceeding the simple case of the original intent of the Single Step Batch Job? 
For this case, I don't think it's exceeding it, but I could be wrong (And I've been wrong many times before ;-) ).   

2. My point is that leaving 5 lines of configuration code to the user is probably "wiser" than taking care of that on the framework side and having to manage 50+ properties and maintain them in the long term. WDYT?
I'm not sure I understand, I thought the properties would be managed by the [DataSourceProperties](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java) associated with the JDBC reader and JDBC writer.   

I look forward to our conversation on this!

spring-cloud-starter-single-step-batch-job/src/main/java/org/springframework/cloud/task/batch/autoconfigure/jdbc/JDBCSingleStepDataSourceAutoConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.cloud.task.configuration.DefaultTaskConfigurer;
+import org.springframework.cloud.task.configuration.TaskConfigurer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Primary;
+
+/**
+ * Establishes the default {@link DataSource} for the Task when creating a {@link DataSource}
+ * for  {@link org.springframework.batch.item.database.JdbcBatchItemWriter}
```
benas, line 33: The sentence mentions `JdbcBatchItemWriter` twice. I guess this line should be `for {@Link org.springframework.batch.item.database.JdbcCursorItemReader}`

cppwfs:
Squashed Merged Polished.  Thanks for the review @benas

Issue #852, opened 2022-07-07
Improve replication friendliness of the MySQL schema


hpoettker:
### Status quo

The tables `TASK_EXECUTION_PARAMS` and `TASK_TASK_BATCH` currently do not have primary keys.

For a single node MySQL deployment, this does not cause any issue. For a setup of multiple nodes with replication, this is however not ideal as it causes performance problems with row-based replication and leads to ambiguity when calculating check sums over tables to validate the integrity of the replication.

To prevent such issues, MySQL 8 offers the system variable [`sql_require_primary_key`](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_sql_require_primary_key). If set to `ON`, it is currently not possible to deploy the Spring Cloud Task schema.

I've opened an analogous issue for Spring Batch: spring-projects/spring-batch#4146.

### Suggestion

It would be nice if primary keys were defined on `TASK_EXECUTION_PARAMS` and `TASK_TASK_BATCH` in the MySQL schema, e.g.

- on `TASK_EXECUTION_ID` and `TASK_PARAM` for `TASK_EXECUTION_PARAMS`,
- on `JOB_EXECUTION_ID` for `TASK_TASK_BATCH` (uniqueness of job execution ids implied by code in `JdbcTaskExecutionDao`).

The table `TASK_SEQ` also has no primary key. As it contains only a single row, this is not a problem in practice. But if the column `UNIQUE_KEY` were to be declared as primary key, this would allow a seamless deployment to a MySQL instance with `sql_require_primary_key=ON`.

hpoettker:
For Spring Batch, I've opened spring-projects/spring-batch#4176. A similar PR, which sets primary keys with minimal invasiveness, is more difficult for Spring Cloud Task.

The problem is that `TASK_PARAM` of `TASK_EXECUTION_PARAMS` is nullable and, thus, cannot be part of a primary key.

I didn't spot a problem in the Spring Cloud Task code with parameters being `null` but doubt that all deployers would react well on a command line parameter being `null`. I'm also not sure what meaning a `NULL` value in this column could have. Should this column actually be non-nullable?

cppwfs:
Hello @hpoettker,
[...]

PR #857, opened 2022-08-25
Adjust SQL statement to allow added columns

hpoettker:
This PR is related to #852.

It adjusts an insert query for the table `TASK_TASK_BATCH` such that it allows additional columns if required by the user.


cppwfs:
LGTM 
Thank you again for another great contribution!

Issue #769, opened 2021-02-19
Allow the user to select the transactionManager to use for TaskRepository


cppwfs:
Currently Task Repositories use `@transactional("springCloudTaskTransactionManager")`.   This should be replaced with a property i.e.
`@transactional("${spring.cloud.task.transaction-manager:transactionManager}")`.  
Also remove springCloudTaskTransactionManager bean.

This is for 3.0.0 release of SCT.

cppwfs:
This was feature was added here: https://github.com/spring-cloud/spring-cloud-task/commit/e4581420b04e872f8aac4cb3ad8bab62bbe9c941

GeniusMagnus:
Recently, I had to make use of the _spring.cloud.task.transaction-manager_ property. On this occasion I observed that class SimpleTaskAutoConfiguration still instantiates a (useless) transaction manager which is not injected into the application context. This could be optimized.

cppwfs:
@GeniusMagnus created [issue](https://github.com/spring-cloud/spring-cloud-task/issues/954) to investigate.

GeniusMagnus:
@cppwfs : Thank you! I actually thought nobody would be interested in my comment as this ticket was closed nearly two years ago. 😏

cppwfs:
@GeniusMagnus thank you for reaching out!   It is much appreciated!

PR #858, opened 2022-08-30
Allow users to select the transaction manager to use in TaskRepository

fmbenhassine:
Allow users to select the transaction manager to use in `TaskRepository`

* Update `@Transactional` attribute to use `spring.cloud.task.transaction-manager` property
* Make `springCloudTaskTransactionManager` bean definition conditional on a missing property

Issue #769


cppwfs:
LGTM !   Thank you for the contribution!!!!!

PR #866, opened 2022-10-03
Updated Task to the latests snapshot of Micrometer [DO NOT MERGE TILL AFTER M5

cppwfs:
Remove the snapshot before merge


cppwfs:
Changes already applied in another PR.

PR #876, opened 2022-11-20
Add runtime hints for `TaskBatchEventListenerBeanPostProcessor`

hpoettker:
This PR adds runtime hints for `TaskBatchEventListenerBeanPostProcessor`, which reflects on fields of `ChunkOrientedTasklet`.

PR #878, opened 2022-11-22
Add conditional runtime hint for `TaskBatchEventListenerBeanPostProcessor`

hpoettker:
This is a follow-up on #876 which led to issues with apps without Spring Batch on the classpath.

@cppwfs Please have another look! It works nicely with my sample apps.

Something I just noticed however: The `TaskBatchExecutionListener` does not work as expected in a native image as the method `onTaskStartup` seems to be compiled away by GraalVM.


cppwfs:
LGTM! Isolating the hint to the bean resolved the previous issue.   Thanks for the contribution!

PR #879, opened 2022-11-22
Let `TaskBatchExecutionListener` implement `TaskExecutionListener`

hpoettker:
When I run a very simple app with Spring Batch wrapped in Spring Cloud Task, the method `TaskBatchExecutionListener::onTaskStartup` seems to be compiled away in a native image as the execution is not set, and a warning is logged accordingly:

> WARN [main] o.s.c.t.b.l.TaskBatchExecutionListener   : This job was executed outside the scope of a task but still used the task listener.

I'm using Spring Boot 3.0.0-RC2 and Spring Cloud 2022.0.0-RC2. When I remove the annotation `@BeforeTask` and let the listener implement the interface `TaskBatchExecutionListener` everything works fine.

It's unclear to me what happens. As `@BeforeTask` is meta-annotated with `@Reflective`, this change should not be necessary.


cppwfs:
I agree.   I'll research it. 
But in the meantime I'd rather implement the interface here instead of the annotation.   
LGTM

PR #887, opened 2023-02-04
Simplify monitoring of job executions

hpoettker:
When `spring.cloud.task.batch.fail-on-job-failure` is set to `true`, started `JobExecution`s are monitored on whether they end in `BatchStatus.FAILED`. This monitoring is performed by database queries on the `JobExecution`s and sleeps to wait until no `JobExecution` is running any more.

In the happy case of `JobExecution`s with `BatchStatus.COMPLETED`, all `JobExecutions` are currently re-queried and a sleep of 5 seconds by default is executed, which may delay application shut-down unnecessarily.

This PR makes the following changes:
- The application runner only sleeps when a `JobExecution` is still running.
- The state of the job executions is only queried for running job executions.

The PR also includes some polishing in `TaskJobLauncherApplicationRunner` and its test class.


cppwfs:
LGTM Thanks for the contribution!

Issue #882, opened 2022-12-02
TaskJobLauncherApplicationRunner overrides default job launcher behavior


aritzbastida:
According to the Spring Cloud Task reference documentation, the property `spring.cloud.task.batch.fail-on-job-failure` can be set to `true `in order to exit with code 1 when a Spring Batch fails. 

I would just expect that implemented with an application event listener subscribed to `JobExecutionEvent`, but I was surprised to find out that switching on this toggle changes the default application runner altogether. Spring Boot's `JobLauncherApplicationRunner` is replaced with `TaskJobLauncherApplicationRunner`, and that overrides internal behavior such as checking previous job executions, incrementing job parameters and the like. Also, unlike `JobLauncherApplicationRunner`, the jobs are monitored with a polling interval, but normally, the JobLauncher implementation should use a `SyncTaskExecutor` in a standalone (task) context. 

Why is all this necessary? And if so, shouldn't this be clearly documented?

cppwfs:
Update TaskJobLauncherApplicationRunner behavior to match JobApplicationRunner behavior.

cppwfs:
Documentation needs to be updated.

cppwfs:
Could you provide the Spring Cloud Task, Spring Batch, and Spring Boot versions where this issue occurred?
Also can you provide a small sample app that exhibits the behavior?  
Thank You!

aritzbastida:
Sure thing. The versions are as follows: Spring Cloud Task 2.4.3, Spring Batch 4.3.6, Spring Boot 2.7.1.

I also attach the sources of a test application. You can turn on/off the property "spring.cloud.task.batch.fail-on-job-failure" in application.yml, as needed.

Summing up, my concerns/questions are:
* `TaskJobLauncherApplicationRunner` is a subclass of `JobLauncherApplicationRunner`, but, in addition to exiting with code 1 when the batch fails, it "overrides" the default behavior altogether (calculating job parameters, monitoring job output, etc.). I don't really understand why two different runners are necessary, and if so, what is the difference between them.
* In case the only expected behavior is exiting with code 1, auto-configuring a default ExitCodeExceptionMapper with @ConditionalOnMissingBean should be enough, in line with the [reference documentation of Spring Cloud Task](https://docs.spring.io/spring-cloud-task/docs/current/reference/html/#features-task-execution-details). 
[...]

PR #888, opened 2023-02-04
Hack JobLauncherApplicationRunner

hpoettker:
This PR relates to my comment https://github.com/spring-cloud/spring-cloud-task/issues/882#issuecomment-1416798769. It's intended to show that a lot of presumeably duplicated code could be removed from `TaskJobLauncherApplicationRunner`.

The PR is very hacky and not intended as a contribution. It's intended to show that a new extension point in `JobLauncherApplicationRunner` would be a solution to #882.


cppwfs:
Thank you for introducing this topic and draft PR!

PR #889, opened 2023-02-08
Inject event publisher into `TaskJobLauncherApplicationRunner`

hpoettker:
Currently, no `ApplicationEventPublisher` is injected into the `TaskJobLauncherApplicationRunner` as in the case of the `JobLauncherApplicationRunner`. Although `TaskJobLauncherApplicationRunner` implements `ApplicationEventPublisherAware` via `JobLauncherApplicationRunner`, the corresponding setter is not invoked as `TaskJobLauncherApplicationRunner` is configured with a `FactoryBean`.

This PR fixes the issue and adds assertions for the correct behavior in `TaskJobLauncherApplicationRunnerTests`.


cppwfs:
LGTM !

Thanks for the contribution!!!!!

PR #892, opened 2023-02-23
Added Spring-boot-starter-jdbc to Task Starter

### LINKED ISSUE #758: Add jdbc-api dependency automatically when cloud-task is selected
In almost every case when a user selects the Spring Cloud Task dependency (cloud-task) the will have to select the JDBC-API dependency as well.
So for this scenario we need to add the JDBC-API dependency anytime a user selects the cloud-task dependency.
This should help users when creating their first Spring Cloud Task Project.

This change will be applied to the task starter.

cppwfs:
The user adds this  dependency is added each time they  implement Spring Cloud task

Resolves #758


cppwfs:
Rebased Merged

PR #903, opened 2023-06-27
SCT no longer generates EPUB docs.

cppwfs:
Remove reference in the documentation.
Resolves #ABC


cppwfs:
We noticed that there is a space before the period.   Can you remove it for us.

cppwfs:
LGTM

Issue #917, opened 2023-09-05
Link to BatchStatus doc points to 4.0.x it should refer to current release


cppwfs:
Replace `4.0.x` in the link with `current`.

The doc to be updated can be found here: 
https://github.com/spring-cloud/spring-cloud-task/blob/main/docs/modules/ROOT/pages/batch.adoc?plain=1#L208

cppwfs:
Updated location of the issue due to migration to Antora.

cppwfs:
Rebased Meged

PR #921, opened 2023-09-12
#917-modified the L208 to have modified to have 'current' in place of '4.0.x' in the link

ksanjeev9211:


ksanjeev9211:
Thank you!!

PR #941, opened 2024-01-04
Fixes #940 Update getting started docs, reduce possible confusion, fix example, …

### LINKED ISSUE #940: The documentation references old initializr and needs to be updated to avoid confusion.
https://docs.spring.io/spring-cloud-task/docs/current/reference/html/#getting-started

1. `In the Dependencies text box, type jdbc and then select the JDBC dependency.`

JDBC?  JDBC API or Spring Data JDBC? Multiple things show up when you type JDBC.

2. `When our sample application runs, Spring Boot launches our HelloWorldCommandLineRunner and outputs our “Hello, World!” message to standard out. The TaskLifecycleListener records the start of the task and the end of the task in the repository.`

3. The example is not using CommandLineRunner, its using ApplicationRunner.

4. The package name in the example doesn't match what we get from Initializr.

5. You don't need to have Maven installed because you get the Maven Wrapper with Initializr.

dashaun:
…update output, with help from Harpreet Singh and Jonas Andersen

### DIFF
--- a/docs/modules/ROOT/pages/getting-started.adoc
+++ b/docs/modules/ROOT/pages/getting-started.adoc
@@ -20,8 +20,7 @@ environment.
 [[getting-started-system-requirements]]
 == System Requirements
 
-You need to have Java installed (Java 17 or better). To build, you need to have Maven
-installed as well.
+You need to have Java installed (Java 17 or better).
 
 [[database-requirements]]
 === Database Requirements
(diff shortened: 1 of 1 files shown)


cppwfs:
LGTM!   Thank you for your contribution!!!

cppwfs:
Rebased, Polished, Squashed, Merged.  Again thank you for your contribution!

